PR #23, opened 2017-09-07
Added an algorithm for finding nth term of fibonacci sequence.

rudrasohan:
Using Divide and Conquer.
This algorithm takes T = O(logn) time which is quite small as compared to the traditional method which takes O(n).
This is my first commit so any constructive criticism is always welcome.

### DIFF
--- a/fibonacci.cpp
+++ b/fibonacci.cpp
@@ -1,0 +1,73 @@
+#include <iostream>
+using namespace std;
+
+
+/*
+	Efficient method for finding nth term in a fibonacci number sequence.
+	Uses Dvide and conquer approach
+	Enter Values from 1
+	Eg-
+	1st term = 0;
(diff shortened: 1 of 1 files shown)


fibonacci.cpp
```diff
@@ -0,0 +1,73 @@
...
+	if(n == 1)
+	{
+		cout<<"Ans: 0"<<endl;
+	}
+	else
+	{	
+		power_rec(n-1);
+		cout <<"Ans :"<<ans[0][1]<<endl;
+	}
+}
```
dynamitechetan: also please leave a blank line at the end of file

rudrasohan:
@dynamitechetan,@sachinarora707 I have done the necessary changes. Please do let me know if something else is required. In the new descriptions, I have also included the core idea of the algorithm.

Issue #24, opened 2017-09-28
Add more algorithms


AnupKumarPanwar:
This repo has less contributions. Please add more algorithms. 

See the structure and README of this repo for reference
https://github.com/TheAlgorithms/Python

goyaljai:
i would like to contribute. thanks

dynamitechetan:
@goyaljai  Sure. Thanks.

goyaljai:
which type of algortihms? and where and how to add?

joealam:
Is there any kind of contribution guide, or at least more information about the desired format of the contributions?

For example: should every algorithm be a single file? should they aim to be runnable without additional dependencies? etc.

shivamgor498:
can we add cocktail sort algorithm??

AnupKumarPanwar:
@joealam A single file is preferred and keep modular approach. Divide you code into function instead of breaking into multiple files and use the general head files for basic algorithms.

Vehemos:
I would like to contribute to this. Thanks!

vishvanath45:
I would like to contribute by adding Dijsktra Algorithm.

shubhammatta:
I would like to add - Disjoint_Set_Union , Range_Minimum_query(Using Sparse Table) , KMP(Knuth–Morris–Pratt) Algorithm , Lowest_Common_Ancestor.

VibhorCodecianGupta:
I'd like to add some sweet functions to Linked Lists and other algorithms like Count Sort, segmented sieve and totient functions. It'd be great if you acknowledged!

arpanjain97:
I'd like to add the following shortest distance algorithms :  
1. Dijkstra
2. Bellman-Ford
3. Floyd-Warshall

nghiaxlee:
Hello, I'd like to add Comb Sort algorithm

rohitchugh9:
I would like to add Dynamic Programming Algorithms.
1. Edit Distance
2. 0/1 Knapsack
3. Egg Dropping Puzzle
4. Subset Sum
5. All Valid Parentheses

rsenwar:
#138 Have a look at this PR, I am looking to add more programs related to Mathematics.

GOVINDDIXIT:
I want to add optimal BST in DP category

anudishjain:
Hi, I would like to add the following Dynamic Programming Algorithms,

- Trapping Rain Water (1D)
- Maximum Product Subarray
[...]

PR #32, opened 2017-10-01
DFS with stack

dcariotti:
As requested in https://github.com/TheAlgorithms/C-Plus-Plus/issues/24 I added a variant of DFS with stack

### DIFF
--- a/Graph/DFS_with_stack.cc
+++ b/Graph/DFS_with_stack.cc
@@ -1,0 +1,53 @@
+#include <iostream>
+#include <list>
+#include <stack>
+
+
+#define WHITE 0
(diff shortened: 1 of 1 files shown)


madhavgaba:
Isn't recursive implementation better than it, in terms of space?

dcariotti:
Well, they're the same thing, more or less. With stack you just use methods push/pop but with recursion you've got a difference in runtime not significant.
The result in output, in general, is different but, if you test the algorithm with a pencil and a paper, you can see the real result (same result with a recursive DFS).

Anyway, the difference in runtime is statistically insignificant.

PR #35, opened 2017-10-04
added binary indexed tree

sb1997:


christianbender:
@sb1997 Can you remove bits/stdc++.h because this is quite Linux specific and slow down the compiler process? In additional can you describe your pull request?

PR #41, opened 2017-10-07
Pull Request

JosephTLyons:
I don't believe C++ allows for variables length arrays, so I implemented a dynamic array with a pointer.  I also cleaned up formatting a tiny bit.

### DIFF
--- a/Sorting/Bubble Sort.cpp
+++ b/Sorting/Bubble Sort.cpp
@@ -6,6 +6,8 @@ using namespace std;
 int main()
 {
 	int n;
+    
+    cout << "Enter an array size: ";
 	cin >> n;
     
     int *arrayPtr = new int[n];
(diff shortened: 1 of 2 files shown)


JosephTLyons:
There are others that seem to be using VLAs as well.. if this pull request is accepted, I will fix the others as well.

JosephTLyons:
This can be closed if there is no interest in accepting the changes.

christianbender:
@JosephTLyons  C++ contains some useful data structures in its standard library. These includes dynamic arrays such as Vectors.

JosephTLyons:
Yes, if you use a vector, you’re fine, if you allocate an array through a pointer, you’re fine, but to my knowledge, you cannot have VLAs outside of these two structures in C++17.

PR #42, opened 2017-10-08
To find the height of tree

ayaanfaiz:
An algorithm to find the height of the tree. I had created a new folder for the Tree so that the repository remains in an organised manner.


christianbender:
@ayaanfaiz  Thanks for contributing. Can you add the files into the directory **Datastructures**?
In this directory you can add a directory **Tree**.

PR #47, opened 2017-10-15
Reorganisation of folders and Updated Readme.md

yeggasd:
Created Searches. Conversions, Data Structures and Others folder similar
to how other languages' repo does it.

Pulled information from other languages' repo to update readme


christianbender:
@yeggasd Can you remove the header  **<bits/stdc++.h>** because that header is quite linux specific.And put in some comments in your code.

PR #48, opened 2017-10-15
Comb Sort algorithm

nghiaxlee:


dynamitechetan:
@nghiaxlee  thanks for the PR 🎉. Please add some comments so that everyone can understand your code easily.

christianbender:
@nghiaxlee Can you remove the header **<bits/stdc++.h>** because this header is quite Linux specific and slow down the compiler-process?

PR #52, opened 2017-10-22
recursive bubble sort

SHAKTISINGH2323:
bubble sort using recursion technique

### DIFF
--- a/Sorting/recursive_bubbleSort.cpp
+++ b/Sorting/recursive_bubbleSort.cpp
@@ -1,0 +1,40 @@
+#include <bits/stdc++.h>
+using namespace std;
+ 
+
+void bubbleSort(int arr[], int n)   //bubble sort FUNCTION
+{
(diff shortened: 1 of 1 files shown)


Sorting/recursive_bubbleSort.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+void bubbleSort(int arr[], int n)   //bubble sort FUNCTION
+{
+    // Base case
+    if (n == 1)
+        return;
+ 
+    
+    for (int i=0; i<n-1; i++)
+        if (arr[i] > arr[i+1])
```
jfsanin: IMHO you can improve this algorithm just looking if there is a swap, if there is not the bubble sort could end in that iteration.

SHAKTISINGH2323: where would be the swap,in the for loop?

PR #54, opened 2017-10-25
Binary Heap and Dijkstra's Algorithm

t0rr3sp3dr0:
Issue #24


christianbender:
@t0rr3sp3dr0  Can you use the class-keyword instead of **struct** ? Because that more C++ like.

PR #56, opened 2017-10-27
Update Kruskals Minimum Spanning Tree.cpp

madhavgaba:
This will work for nodes>1000 for which the current code fails


christianbender:
@madhavgaba  Can you remove the header **"bits/stdc++.h"**? Because this header is quite linux specific and slow down the compile time.

madhav30:
As per my experience, "bits/stdc++.h" works well on all operating systems. In competitive programming we are more concerned with run-time and not the compile time, therefore I think,  it need not be replaced.

PR #61, opened 2017-10-28
Longest Palindromic Subsequence

lo-0l:


christianbender:
@lo-0l  Can you remove the include **bits/stdc++.h**  because this header is quiet Linux specific and slow down the compilation process.

PR #62, opened 2017-10-28
Coin Change Problem

lo-0l:
Added Coin Change Problem in DP section


christianbender:
Can you write blocks in your code?
for instance
``` c++
for (int i = 0; i < 10; i++)
{
    // some code
}

```

instead of 

``` c++
for (int i = 0; i < 10; i++) cout <<  i*i;

```

and use either **cout** or **printf**

I tested it now and I have get a memory access error.
bildschirmfoto_2017-12-23_18-05-34

PR #63, opened 2017-10-28
Subset Sum Problem

lo-0l:
Subset Sum Problem added to DP


christianbender:
@lo-0l 

Can you remove **bits/stdc++.h** and use the appropriate headers? Because bits/stdc++.h is a big header and unnecessary  slow down the compilation-process.

And use either **printf** or **cout**

PR #64, opened 2017-10-28
Create Max1DRangeSum.cpp

MahmoudTarek97:


christianbender:
@MahmoudTarek97  Please comment your code and remove the header **<bits/stdc++.h>**. Because this header is quite linux specific.

PR #67, opened 2017-12-17
Create Primality Test.cpp

hegdenaveen1:
A simple and efficient Primality Test Algorithm which was missing.


christianbender:
@hegdenaveen1 

Can you remove **bits/stdc++.h** and use the appropriate headers? Because bits/stdc++.h is a big header and unnecessary slow down the compilation-process.

PR #68, opened 2017-12-18
Add BigInt Class

hegdenaveen1:
BigInt Class can be used in Competitive Contests which may require the use of BigIntegers to be used in C++


christianbender:
@hegdenaveen1  

Can you remove **bits/stdc++.h** and use the appropriate headers? Because bits/stdc++.h is a big header and slow down the compilation-process.
In additional, can you use the **class** keyword instead of **structs**?

PR #75, opened 2017-12-25
Updated logic of fibonacci.cpp

hegdenaveen1:
Implemented an easier to implement and faster algorithm than the already present matrix exponentiation method to find nth fibonacci number.

This is faster than matrix exponentiation method as it is a DP form of matrix exponentiation.

Faster than O(logn)


hegdenaveen1:
source of algorithm  https://www.geeksforgeeks.org/program-for-nth-fibonacci-number/

christianbender:
cool!
You can change **long long** in the for-loop at the main function into **int**.

PR #83, opened 2017-12-26
Added String Fibonacci

hegdenaveen1:
This programme uses String Addition to find Nth fibonacci.


christianbender:
@hegdenaveen1  You can remove that **std::** because you use **using namespace std**

hegdenaveen1:
done.

PR #93, opened 2018-04-21
Edits #include directives

linus-xuzixuan:
My computer doesn't seem to have the stdc++.h, so I tried changing it to the two headers really used. It seems to be solving the problem.


linus-xuzixuan:
I tried to download this file from Google, but I failed to get write access to /usr/include/c++/4.2.1/bits to put the header in it. (I don't know why. Even sudo chmod 777 didn't work.)

christianbender:
@linus-xuzixuan Thanks. Thats to the ground why we don't accept the header ```<bits/stdc++.h>```. Because this header is quite linux specific.

PR #94, opened 2018-04-27
Odd_Even Sort

tosirap:


christianbender:
Removed the file ```OddEven Sort.cpp```.

Sorting/OddEven Sort.cpp
```diff
@@ -0,0 +1,68 @@
...
+			}
+		}
+	}
+}
+
+void show(int A[], int size)
+{
+	int i;
+	for (i = 0; i < size; i++)
+	cout<<A[i]<<"\n";
```
christianbender: Please put in spaces between the operators like ```<<```.

christianbender: Please ident this line (line number 42)

Sorting/OddEven Sort.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+
+int main()
+{
+	int size;
+	cout<<"\nEnter the number of elements : ";
+
+	cin>>size;
```
christianbender: You can avoid many lines. Use simply ```std::vector```

Sorting/OddEven Sort.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+	int arr[size];
```
christianbender: See comment above. Use ```std::vector```

Sorting/OddEven_Sort.cpp
```diff
@@ -0,0 +1,68 @@
+/* C++ implementation Odd Even Sort */
+#include<iostream>
+using namespace std;
+
```
christianbender: The same as my suggestions above.

PR #99, opened 2018-08-29
Added FenwickTree For Range Query

DDullahan:
FenwickTree also named as Binary-Indexed-Tree. It organizes data using binary characteristics.
Point Update and Range Query can all be done with O( log (n) ) , and has less hidden constant than Segement-Tree.

What's more, Range Update and Point Query can be done with FenwickTree , I will pr as soon as i finish it, also ST , matrix quick pow .

I'm glad if you can give me some suggestions with my first pr   :-)

### DIFF
--- a/Range queries/FenwickTree.cpp
+++ b/Range queries/FenwickTree.cpp
@@ -1,0 +1,56 @@
+#include<iostream>
+using namespace std;
+
+/**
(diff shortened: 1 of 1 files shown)


DDullahan:
plz check my PR

PR #107, opened 2018-10-01
added knight tour to backtracking and longest common string to dynamic

nikhilarora068:
added 2 new algorithms please have a look


Backtracking/knight-tour.cpp
```diff
@@ -0,0 +1,59 @@
+#include<bits/stdc++.h>
```
ashwek: please don't use **bits/stdc++.h** because this is quite Linux specific and slows down the compiler process.

PR #108, opened 2018-10-01
Create Ugly Numbers

gholakemohit:
Dynamic Programming Problem

PR #110, opened 2018-10-01
Create Subset Sum Problem

gholakemohit:


ashwek:
Original source [geeksforgeeks](https://www.geeksforgeeks.org/subset-sum-problem-dp-25/)
Refer to PR #108

PR #113, opened 2018-10-01
Added Geometry library

JoaoMLima:


nikhilkala:
The library needs to be in a directory.

PR #114, opened 2018-10-02
Added Bitonic Sort

Aruj-Sharma:


ashwek:
You could have implemented the code by yourself instead of copying it from [geeksforgeeks](https://www.geeksforgeeks.org/bitonic-sort/), or at least could have added the source in the comments.

PR #118, opened 2018-10-02
add: sort technique using vectors

piyushagru:


ashwek:
@piyushagru Please move the file to **_Others_** directory, and please rename the file so the name reveals what the programs is doing.

PR #119, opened 2018-10-03
queue template

aksharanigam1112:


ashwek:
@aksharanigam1112 you have added 2 copies of **queue_template.cpp** please remove 1. And please move **stack_template.cpp** to the **_Datastructure_** folder

Datastructures/queue_template.cpp
```diff
@@ -0,0 +1,104 @@
...
+        else
+            return false;
+    }
+    void Insert(t x)
+    {
+        if(!isFull())
+        {
+            arr[e] = x;
+            e++;
+            s++;
```
ashwek: Your **Queue would never be full** because you are **updating the s** _(size)_ too.

Datastructures/queue_template.cpp
```diff
@@ -0,0 +1,104 @@
...
+    {
+        return s;
+    }
+    t Front()
+    {
+        return arr[f];
+    }
+    t Back()
+    {
+        return arr[e];
```
ashwek: before returning Front / Back you should check whether Queue is empty / not.
Back (rear) element should be **arr[e-1]**

stack_template.cpp
```diff
@@ -0,0 +1,76 @@
...
+        return (top+1);
+    }
+    bool isEmpty()
+    {
+        if(top==-1)
+            return true;
+    }
+    bool isFull()
+    {
+        if (top == capacity)
```
ashwek: it should be ```top+1 == capacity```

vector_template.cpp
```diff
@@ -0,0 +1,119 @@
...
+        {
+            t n = v[0];
+            t*temp = new t[s];
+            for(int i=0;i<s-1;i++)
+            {
+                temp[i] = v[i+1];
+            }
+            v = temp;
+            s--;
+            capacity-=2;
```
ashwek: deleting an element should change the size only, not the capacity.

PR #124, opened 2018-10-05
Added Edmonds–Karp Max-Flow-Min-Cut Algorithm.

metamemelord:


Graph/Edmonds–Karp-Algorithm.cpp
```diff
@@ -0,0 +1,158 @@
+/*
+# -*- coding: utf-8 -*-
+"""
+@author: MetaMemeLord-
+"""
+*/
+
+#include<bits/stdc++.h>
```
ashwek: please don't use **bits/stdc++.h** because this is quite Linux specific and slows down the compiler process. _(Specified in [README](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/README.md))_

PR #125, opened 2018-10-06
An algorithm to encrypt and decrypt the passwords..

rahulgoyal911:


ashwek:
@rahulgoyal911 thank you for contributing. Please rename the **encryption.cpp** file to the encryption method you are implementing & move the file to the **_Others/_** directory.
Please resolve merge conflict.

PR #128, opened 2018-10-07
Ayaan bhai kaisa h tu?

kartikgandhi:
bhai bhul hi gya tu call b nhi krta

### DIFF
--- a/Trees/Top view.cpp
+++ b/Trees/Top view.cpp
@@ -1,0 +1,83 @@
+// C++ program to print top 
+// view of binary tree
+#include<bits/stdc++.h>
+using namespace std;
+ 
+//Structure of binary tree
+struct Node
+{
+    int data;
+    struct Node* left, * right;
+};
+ 
+// function should print the topView of 
+// the binary tree
+void topView(struct Node *root)
+{
+    if (root == NULL)
+        return;
+ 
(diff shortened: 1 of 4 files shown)


Trees/Valid BST or not.cpp
```diff
@@ -0,0 +1,60 @@
+// C++ program to check if a given tree is BST.
+#include <bits/stdc++.h>
```
ashwek: please don't use **bits/stdc++.h** because this is quite Linux specific and slows down the compiler process. _(Specified in [README](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/README.md))_

Trees/Top view.cpp
```diff
@@ -0,0 +1,83 @@
+// C++ program to print top 
+// view of binary tree
+#include<bits/stdc++.h>
```
ashwek: please don't use **bits/stdc++.h** because this is quite Linux specific and slows down the compiler process. _(Specified in [README](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/README.md))_

ashwek:
@kartikgandhi thank you for contributing, but we already have an implementation of binary tree. It would be great if you could add functionalities / improve the already existing implementation.
Closing PR.

PR #133, opened 2018-10-11
Trie Tree

shubhamguptaji:


Datastructures/TrieTree.cpp
```diff
@@ -0,0 +1,97 @@
...
+ string, if it is then just sets the endofword to false, else 
+ removes the given string*/
+bool deleteString (trie *root, char* str, int index)
+{
+    if (index == strlen(str))
+    {
+        if (!root->isEndofWord)
+            return false;
+        root->isEndofWord = false;
+        for (int i = 0; i < 26; i++)
```
kvedala, line 64: @ms10398 Can you please explain about these two lines?
```cpp
for (int i = 0; i < 26; i++)
    return false;
```
As I understand, this loop will execute exactly once and will **always** return `false` and never `true`.
Can you fix that? If you can document the code, that would be much helpful. Thanks.

PR #137, opened 2018-10-15
Program for Huffman Coding in C++

yashmunoth:


Greedy Algorithms/huffman.cpp
```diff
@@ -0,0 +1,108 @@
+// C++ program for Huffman Coding 
+#include <bits/stdc++.h>
```
ashwek: please don't use **bits/stdc++.h** because this is quite Linux specific and slows down the compiler process. _(Specified in [README](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/README.md))_

PR #141, opened 2018-10-16
Update Decimal To Binary.cpp

rsenwar:
Improved code with proper output and User friendly


ashwek:
Thank you for contributing. But similar changes were already made my another PR (#97)
Closing the PR.

PR #149, opened 2018-10-28
Added new cpp programs

Viz38:
The following programs are added
1. Program to find the area of a given shape (area_of_shape.cpp)
2. Program to find the character count in a given string (char_counter.cpp)
3. Program to calculate a bill ( bill.cpp)
4. Program to do simple calculations (calculator.c)
5. Program to store employee details (employee.cpp)
6. Program to check odd/even/prime of a given number (odd_eve_prime.cpp)
7. Program to convert a given string to proper case (proper_case.cpp)
8. Program to reverse a string without using library function (reverse.cpp)
9. Program to sum of all odd/even numbers in a given limit ( sum_of_odd_eve.cpp)

P.S I've tried to explain as much as possible, the programs are easy enough.

Cheers *thumbsup*

### DIFF
--- a/area_of_shape.cpp
+++ b/area_of_shape.cpp
@@ -1,0 +1,50 @@
+#include<iostream>
+using namespace std;
+
+
+  int main()
+   {
+    int side1,side2,option,flag=0;
+    float area=0;
+    while(!flag)
+    {
+     system("cls");
+     cout<<"\n********FIND THE AREA********\n\n";
+     cout<<"Enter the shape\n";
+     cout<<"1.Circle\n2.Square\n3.Rectangle\n4.Exit\n";
+     cout<<"Choice :";
+     cin>>option;
+
+     switch(option)
+      {
+       //circle
+       case 1:cout<<"Enter the radius of the circle in centimeters\n";
+	      cin>>side1;
+	      area=3.14*side1*side1;
+	      cout<<"The area of the circle = "<<area<<endl;
+	      system("pause");
+	      break;
+       //square
+       case 2:cout<<"Enter the side of the square in centimeters\n";
+	      cin>>side1;
+	      area=side1*side1;
+	      cout<<"The area of the square = "<<area<<endl;
+	      system("pause");
+	      break;
+       //rectangle
+       case 3:cout<<"Enter the length and breadth of the rectangle in centimeters\n";
+	      cin>>side1>>side2;
+	      area=side1*side2;
+	      cout<<"The area of the rectangle ="<<area<<endl;
+	      system("pause");
+	      break;
+       case 4:exit (0);
+        break;
+       //unspecified shape
+       default:cout<<"Invalid Operator\n";
+           system("pause");
+	       break;
+      }
+    }
+   return 0;
+  }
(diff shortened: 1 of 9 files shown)


area_of_shape.cpp
```diff
@@ -0,0 +1,50 @@
...
+using namespace std;
+
+
+  int main()
+   {
+    int side1,side2,option,flag=0;
+    float area=0;
+    while(!flag)
+    {
+     system("cls");
```
ashwek: **cls** works only on windows machines. It wont work on other OS.

area_of_shape.cpp
```diff
@@ -0,0 +1,50 @@
...
+     cin>>option;
+
+     switch(option)
+      {
+       //circle
+       case 1:cout<<"Enter the radius of the circle in centimeters\n";
+	      cin>>side1;
+	      area=3.14*side1*side1;
+	      cout<<"The area of the circle = "<<area<<endl;
+	      system("pause");
```
ashwek: **pause** command would work only on windows machines.
you may replace _system("pause")_ with **getchar()**.

calculator.c
```diff
@@ -0,0 +1,46 @@
...
+	ansr=num1*num2;
+      //addition
+      else if(oprtr=='+')
+	ansr=num1+num2;
+      //substraction
+      else if(oprtr=='-')
+       ansr=num1-num2;
+      //division
+      else if(oprtr=='/')
+       ansr=num1/num2;
```
ashwek: before dividing please check if ```num2 != 0``` because dividing by 0 will crash the program.

char_counter.cpp
```diff
@@ -0,0 +1,44 @@
+#include<conio.h>
```
ashwek: Please dont use conio.h. conio.h is a C header file used in old MS-DOS compilers. It's platform specific. It is not supported on linux & lacks compatibility with compilers like Gcc.

char_counter.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+//progrram to count the number of alphabets, vowels, digits and words in a string
+
+  int main()
+   {
+    system("cls");
+    int i,alphacount=0,vowelcount=0,degitcount=0,wordcount=0;
+    char a[50];  //storing the string
+    cout<<"\tEnter a string to count\n\t";
+    gets(a);
```
ashwek: [Why is the gets function so dangerous that it should not be used? - StackOveflow](https://stackoverflow.com/questions/1694036/why-is-the-gets-function-so-dangerous-that-it-should-not-be-used)

odd_eve_prime.cpp
```diff
@@ -0,0 +1,48 @@
...
+   system("cls");
+   int num,flago=0,flagp=0,flage=0,flag=0;
+   char choice[4];
+   cout<<"Enter the number\n";
+   cin>>num;
+
+   do
+    {
+     //checking prime
+     for(int i=0;i<num/2;i++)
```
ashwek: in the 1st iteration it will perform **_num %_ 0** crashing the program.

ashwek:
Please create a new directory **Basics** & move these 9 programs in it.

PR #150, opened 2018-10-29
Shift-Or

t0rr3sp3dr0:
Issue #24

### DIFF
--- a/String Searching/shift_or.cpp
+++ b/String Searching/shift_or.cpp
@@ -1,0 +1,78 @@
+#include <bits/stdc++.h>
+#define endl '\n'
+
+using namespace std;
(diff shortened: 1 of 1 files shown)


String Searching/shift_or.cpp
```diff
@@ -0,0 +1,78 @@
+#include <bits/stdc++.h>
```
ashwek: please don't use **bits/stdc++.h** because this is quite Linux specific and slows down the compiler process.

String Searching/shift_or.cpp
```diff
@@ -0,0 +1,78 @@
...
+#define endl '\n'
```
christianbender: Why you define ```endl``` with ```'\n'```? This statment is contain in C++ (namespace std)

PR #151, opened 2018-10-29
Boyer-Moore

t0rr3sp3dr0:
Issue #24

### DIFF
--- a/String Searching/boyer_moore.cpp
+++ b/String Searching/boyer_moore.cpp
@@ -1,0 +1,113 @@
+#include <bits/stdc++.h>
+#define endl '\n'
+
+using namespace std;
+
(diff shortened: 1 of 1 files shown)


String Searching/boyer_moore.cpp
```diff
@@ -0,0 +1,113 @@
+#include <bits/stdc++.h>
```
ashwek: please don't use **bits/stdc++.h** because this is quite Linux specific and slows down the compiler process.

String Searching/boyer_moore.cpp
```diff
@@ -0,0 +1,113 @@
...
+#define endl '\n'
```
christianbender: You don't need define this as präprocessor directive. Its contain in C++ (namespace std)

PR #152, opened 2018-10-29
Wu-Manber

t0rr3sp3dr0:
Issue #24


String Searching/wu_manber.cpp
```diff
@@ -0,0 +1,96 @@
+#include <bits/stdc++.h>
```
ashwek: please don't use **bits/stdc++.h** because this is quite Linux specific and slows down the compiler process.

PR #155, opened 2018-10-30
Added and algorithm to check if a number is palindrome

gopikrishnanrmg:
Added the algorithm to convert number to string and check if it's a palindrome.

### DIFF
--- a/Palindromeofnumber.cpp
+++ b/Palindromeofnumber.cpp
@@ -1,20 +1,23 @@
-#include<bits/stdc++.h>
+#include <iostream>
+#include <algorithm>
+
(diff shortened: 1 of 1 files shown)


gopikrishnanrmg:
Added this algorithm to check if number is palindrome.

Palindromeofnumber.cpp
```diff
@@ -0,0 +1,20 @@
+#include<bits/stdc++.h>
```
ashwek: please don't use **bits/stdc++.h** because this is quite Linux specific and slows down the compiler process.

PR #156, opened 2018-10-31
Implemented sort for string

Deepak-j-p:
Implemented a counting sort for string in cpp


Sorting/CountingSortString.cpp
```diff
@@ -0,0 +1,44 @@
+// C++ Program for counting sort 
+#include <bits/stdc++.h>
```
ashwek: please don't use **bits/stdc++.h** because this is quite Linux specific and slows down the compiler process.

PR #157, opened 2018-10-31
Add files via upload

kuldeepdadhich:


numeric_keypad_problem.cpp
```diff
@@ -0,0 +1,76 @@
+#include <bits/stdc++.h>
```
ashwek: please don't use **bits/stdc++.h** because this is quite Linux specific and slows down the compiler process.

ashwek:
@kuldeepdadhich please add comments describing what the problem is & what the code does.

PR #159, opened 2018-10-31
Added breadth first search

Tarkjot:


Search/bfsnew.cpp
```diff
@@ -0,0 +1,58 @@
...
+int end=0;
+void BFS();
+void push();
+void pop();
+
+int main()
+{
+	for(int i=0;i<5;i++)
+	{
+		cin>>a[i];
```
ashwek: Graph is a pair of _set of vertices & edges_.
What does **a** store? Vertices or Edges??
You should take both _(vertices & edges)_ as input.

Search/bfsnew.cpp
```diff
@@ -0,0 +1,58 @@
...
+	}
+	cout<<"\nPrinting in bfs order";
+	BFS();
+	
+	return 0;
+ } 
+ void BFS()
+ {
+
+ 	queue[0]=a[root];
```
ashwek: you have not initialized **root** so it may take any **random junk value** as its initial value. Assume **root** take 1583 as its initial value, then _a[**1583**]_ may crash the program because size of **a** is 5.

[learncpp.com - uninitialized variables & undefined behavior](https://www.learncpp.com/cpp-tutorial/uninitialized-variables-and-undefined-behavior/)

Search/bfsnew.cpp
```diff
@@ -0,0 +1,58 @@
...
+ 	end++;
+ 	while(front<5)
+ 	{
+ 		pop();
+ 		push();
```
ashwek: in BFS we don't just pop / push values into the queue.
We pop _(from front)_ the nodes / vertices that have been visited.
& push the nodes / vertices that are adjacent to the node in the front of the queue.

ashwek:
@Tarkjot Your program seems to be conceptually incorrect. Closing the PR.

PR #167, opened 2018-11-10
DP program: Catalan numbers

rafi007akhtar:
# Program for Catalan numbers

## Description

A classic DP problem that is often asked in coding contests and interviews is solving for the nth Catalan number.  Catalan numbers have numerous applications, like possible number of expressions with correctly matched parentheses, possible number of BSTs with `n` keys, and so many more. [[Applications of Catalan Numbers - GeeksforGeeks]](https://www.geeksforgeeks.org/applications-of-catalan-numbers/)

## In my program

Using the technique of **tabulation** in dynamic programming, I have written a program that computes and prints the first `n+1` Catalan numbers, where `n` is given by the user. 

Although Catalan numbers have a factorial-based definition, the definition I used was a recursive one. This definition is written in the program as comments (shown below), and also available in its [Wikipedia Page](https://en.wikipedia.org/wiki/Catalan_number#Properties).
```cpp
* A Catalan number satifies the following two properties:
 * C(0) = C(1) = 1; C(n) = sum(C(i).C(n-i-1)), from i = 0 to n-1
```

## Why should this program be included in this repository

- It is a classic DP problem that was thus far missing from this repository.
- Satisfies all the requirements in the [README](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/README.md).
  * [x] Directory structure followed.
  * [x] This PR is fully described.
  * [x] `bits/stdc++.h` is not used; instead I have used the classic `iostream`.
  * [x] I have commented the code as much as I could.
  * [ ] Didn't need `struct` or `class`, so didn't use.
  * [x] Supplied a sample test case in the comments (as I didn't want to put any in the `main` function).
- Solves the problem correctly using a classic DP technique,

## Ending notes

If there is any mistake in my program, please notify in this PR conversation, and I shall try to improve them.

Looking forward to seeing this program in the repository!

### DIFF
--- a/Dynamic Programming/Catalan-Numbers.cpp
+++ b/Dynamic Programming/Catalan-Numbers.cpp
@@ -1,0 +1,79 @@
+/** Print all the Catalan numbers from 0 to n, n being the user input.
+
+ * A Catalan number satifies the following two properties:
+ * C(0) = C(1) = 1; C(n) = sum(C(i).C(n-i-1)), from i = 0 to n-1
+ * Read more about Catalan numbers here:
+    https://en.wikipedia.org/wiki/Catalan_number
+ */
+
+#include <iostream>
+using namespace std;
+
+int *cat; // global array to hold catalan numbers
+
+unsigned long int catalan_dp(int n)
+{
+	/** Using the tabulation technique in dynamic programming,
+        this function computes the first `n+1` Catalan numbers
+
+        Parameter
+        ---------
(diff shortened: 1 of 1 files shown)


ashwek:
@rafi007akhtar Program works fine for inputs less than 20.
For n = 20, Catalan number = 6564120420 (overflows int's capacity). You can use **unsigned int** to extend the range.
Accepting PR. :+1:

PR #171, opened 2018-11-20
Added Topological Sorting

AbhiY98:
Added an implementation of Topological Sort for Directed Graphs.

### DIFF
--- a/Graph/Topological-Sort.cpp
+++ b/Graph/Topological-Sort.cpp
@@ -27,18 +27,19 @@ void topological_sort() {
   reverse(ans.begin(), ans.end());
 }
 int main(){
+  cout << "Enter the number of vertices and the number of directed edges\n";
   cin >> n >> m;
   int x , y;
   G.resize(n , vector<int>());
   for(int i = 0 ; i < n ; ++i) {
     cin >> x >> y;
-    x-- , y--; //to convert 1-indexed to 0-indexed
+    x-- , y--; // to convert 1-indexed to 0-indexed
     G[x].push_back(y);
   }
   topological_sort();
   cout << "Topological Order : \n";
   for(int v : ans) {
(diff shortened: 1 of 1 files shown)


Graph/Topological-Sort.cpp
```diff
@@ -0,0 +1,45 @@
...
+  visited.assign(n, false);
+  ans.clear();
+  for (int i = 0; i < n; ++i) {
+    if (!visited[i])
+      dfs(i);
+  }
+  reverse(ans.begin(), ans.end());
+}
+int main(){
+  cin >> n >> m;
```
ashwek: please add ```cout<<"enter number of vertices and edges = ";``` or something similar so the user knows what input has to be given.

Graph/Topological-Sort.cpp
```diff
@@ -0,0 +1,45 @@
...
+  G.resize(n , vector<int>());
+  for(int i = 0 ; i < n ; ++i) {
+    cin >> x >> y;
+    x-- , y--; //to convert 1-indexed to 0-indexed
+    G[x].push_back(y);
+  }
+  topological_sort();
+  cout << "Topological Order : \n";
+  for(int v : ans) {
+    cout << v << ' ';
```
ashwek: if you are _converting 1-indexed to 0-indexed_ after input, you should _convert 0-indexed to 1-indexed_ before output too. _(User will input **1** vertex but will see **0** in output)_

Graph/Topological-Sort.cpp
AbhiY98: Fixed !

Graph/Topological-Sort.cpp
AbhiY98: Fixed

PR #173, opened 2018-12-29
manacher's algorithm to find palindromic string

God-Hand:
manacher's algorithm to find palindromic string in linear time complexity.

### DIFF
--- a/Manacher's-algorithm.cpp
+++ b/Manacher's-algorithm.cpp
@@ -1,0 +1,62 @@
+// Maximum length palindromic string
+#include<iostream>
+#include<algorithm>
+#include<vector>
+
+using namespace std;
+
+// function to calculate palindromic length with respect to position character at i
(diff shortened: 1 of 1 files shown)


Manacher's-algorithm.cpp
```diff
@@ -0,0 +1,62 @@
...
+// function to calculate palindromic length with respect to position character at i
+int longestPalindrome(int center, int i, string str){
+    if(center-i==-1 || center+i==str.length() || str[center-i]!=str[center+i]){
+        return 0;
+    }
+    if(str[center-i]=='|'){
+        return longestPalindrome(center, i+1, str);
+    }
+    return 1+longestPalindrome(center, i+1, str);
+}
```
tapaswenipathak: I think you are doing a brute force implementation. This should be part of `Manacher` function.

God-Hand: No, it's not.

PR #177, opened 2019-01-09
Conceptually_correct_Insertion Sort.cpp

aryancoder280:
There was a silly mistake that beginners usually do.

### DIFF
--- a/Sorting/Insertion Sort.cpp
+++ b/Sorting/Insertion Sort.cpp
@@ -5,29 +5,47 @@ using namespace std;
 
 int main()
 {
+	/*int n;
+      int Array[n];*/
 	int n;
+	cin>>n;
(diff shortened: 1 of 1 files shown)


Sorting/Insertion Sort.cpp
```diff
@@ -5,29 +5,47 @@ using namespace std;
...
+		*/
+		while(j>=0&&temp<Array[j]) //correct approach:)
+		{
+			j--;
+		}
+		int k;
+		for( k=i;k>j+1;k--)
+		{
+			Array[k]=Array[k-1];
+		}
```
ashwek: isn't it inefficient to use a completely different loop to move elements by 1 position when it could have been done in the while loop itself ?

PR #182, opened 2019-02-23
Add code for computing discrete fourier transfrom

ram-nad:
Discrete Fourier Transform is used for computing polynomial multiplication in O(nlogn) time.


ram-nad:
Thank You. I will do the changes.
I think my description is correct. FFT is an algorithm used for computing DFT. So I wrote it.

PR #183, opened 2019-03-13
Updated remove() and added new showSize() function

ojasiiitd:
**remove()** function was made a little **simpler** and another useful function, **showSize()**, which **returns the size** of the linked list was added.

### DIFF
--- a/Data Structure/Linked List.cpp
+++ b/Data Structure/Linked List.cpp
@@ -3,123 +3,140 @@ using namespace std;
 
 struct node
 {
-	int val;
-	node *next;
+    int val;
+    node *next;
 };
 
 node *start;
 
 void insert(int x)
 {
-	node *t=start;
-	if (start!=NULL)
-	{
-			while(t->next!=NULL)
-		{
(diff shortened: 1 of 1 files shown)


Data Structure/Linked List.cpp
```diff
@@ -3,123 +3,140 @@ using namespace std;
...
+    else if( start->val == x ){
+        node *temp = start;
+        start = start->next;
+        delete temp;
+        return ;
+    }
+
+    node *temp = start;
+
+    while( temp != NULL && temp->next->val != x ){
```
ashwek: suppose there is only 1 node in the Linked list, ``temp->next->val`` will cause error
because ``temp->next`` will be ``NULL``

ashwek:
@ojasiiitd 
Suppose the linked list contains only 1 element, that is **10**.
``Line 130 :`` i give **50** as input. _(50 is not present in the linked list)_,
then ``Line 41`` would be **False**. Hence ``temp->next->val`` on line 50 will be executed & cause error.

PR #200, opened 2019-05-30
add prims algorithm

raj09utkarsh:
Added Prims Algorithm for Minimum Spanning Tree. 
It finds the cost of the the minimum spanning tree.


kvedala:
Can you check how the contribution would be different from an [existing algorithm](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/graph/prim.cpp) in the repo? Thanks

PR #203, opened 2019-06-05
Create Digraph.cpp

sinkyoungdeok:
Create Digraph.cpp

### DIFF
--- a/Data Structure/Digraph.cpp
+++ b/Data Structure/Digraph.cpp
@@ -1,0 +1,51 @@
+#include <iostream>
+#include <vector>
+using namespace std;
+
+struct Edge {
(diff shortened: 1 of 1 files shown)


kvedala:
@fhlasek can you confirm if this code is a duplicate of existing in the repo or can it be considered a good first issue?

fhlasek:
@kvedala I believe we don't yet have a standalone graph data structure implementation checked in the repo. This is indeed something I wanted to create (see #1004) as a library that can be included and reused in graph algorithm. I have a draft at my computer that's probably already in a better state than this PR, but nothing checked in yet - intended to work on it during the weekend.

PR #205, opened 2019-06-05
Create graph.cpp

sinkyoungdeok:
Create graph.cpp


kvedala:
closing this. Seems like a repeat of #203 and we have an entire folder for graph algorithms

PR #206, opened 2019-06-05
Create BinaryTree.cpp

ljw322:
This is the BinaryTree.cpp

please request and check. Thank U : )


ashwek:
@ljw322 thank you for contributing, but we already have an implementation of binary tree. It would be great if you could add features / improve the existing code.
Closing PR.

PR #208, opened 2019-06-06
Add Convex Hull Problem

chaitanyarahalkar:
Jarvis algorithm implemented to solve convex hull problem.


kvedala:
Is this the same as [this file](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/geometry/jarvis_algorithm.cpp) in the repo?

PR #219, opened 2019-06-07
Create Knight_travel.cpp

sinkyoungdeok:


kvedala:
This is a good alternative implementation for [this code](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/backtracking/knight_tour.cpp). but needs to be updated per repo standards

PR #221, opened 2019-06-07
add abs in Cstdlib header

sinkyoungdeok:


michLab:
pt., 7 cze 2019, 15:39 użytkownik sinkyoungdeok <notifications@github.com>
napisał:

ashwek:
Thank you for contributing, but this method is already defined in the C++ libraries, re-implementing it may be of no use unless there is huge improvement in the code.
Closing PR.

PR #242, opened 2019-06-07
Create log10.cpp

ljw322:


ashwek:
Thank you for contributing, but this method is already defined in the C++ libraries, re-implementing it may be of no use unless there is huge improvement in the code.
Closing PR.

PR #274, opened 2019-07-07
Update Armstrong Number.cpp

travellers:
So any positive integer can be checked.


Dynamic Programming/Armstrong Number.cpp
```diff
@@ -1,20 +1,47 @@
 //program to check whether a number is an armstrong number or not
-#include<iostream.h>
+#include<iostream>
 #include<Math.h>
```
ashwek: please change **Math.h** to **math.h**

PR #287, opened 2019-08-20
Interpolation search method

lakshika1064:
This algorithm is some what better than the binary search .It tries to follow the way we search a word in the dictionary , We know in advance that in case the word starts with letter "a" like "alone" for instance , so we know that it should be somewhat at the beginning .This is because we the order of the letters. somewhere we intuitively know that the words are dispersed equally.These facts are enough to realize that the binary search can be a bad choice . So we use this algorithm to improve the binary search


Search/interpolation_search.cpp
```diff
@@ -0,0 +1,41 @@
...
+		}
+	if(arr[low]== value)
+		return low;
+	return 0;
+}
+
+int main() 		//main function
+{
+	int n,value,array[100];
+	cout<<"Enter the size of array : ";
```
ashwek: please specify that the size of the array should be <=100 or dynamically allocate memory according to the input size.

cclauss:
Automated tests are failing
https://github.com/TheAlgorithms/C-Plus-Plus/pull/287/checks?check_run_id=382193607

Search/interpolation_search.cpp
```diff
@@ -1,40 +1,37 @@
...
-        cout<<"Enter the value you want to search : ";
-        cin>>value;
-        re=search(array,value,n);
-
-        if(re==0)
-                cout<<"Entered value is not in the array"<<endl;
+        int n ,value ,array[100] ,re;
+        cout << "Enter the size of array(less than 100) : ";
+        cin >> n;
+        cout << "array in ascending (increasing) order : " << endl;
```
cclauss: Need to add __std::__ like this:
```python
         std::cout << "Enter the size of array(less than 100) : ";
         std::cin >> n;
         std::cout << "array in ascending (increasing) order : " << std::endl;
```

lakshika1064: Ohk but why is it necessary

cclauss: https://stackoverflow.com/questions/1265039/using-std-namespace
TL;DR it can eliminate hard to find bugs where someone defines xyz that there is also a std::xyz

Search/interpolation_search.cpp
```diff
@@ -9,7 +9,8 @@ int search(int arr[], int value, int len) {
             mid = (low + ((value-arr[low])*(high-low)) / (arr[high]-arr[low]));
             if (arr[mid] > value)
                      high = mid-1;
```
cclauss: Please remove line 2: `using namespace std;`

cclauss:
Lower case `s` on the directory name please.

1 files contain uppercase characters:
Search/interpolation_search.cpp

PR #290, opened 2019-09-11
quick sort using randomised partition

mouar:
this sorting is faster than other sorting

### DIFF
--- a/randomised_quicksort.cpp
+++ b/randomised_quicksort.cpp
@@ -1,0 +1,59 @@
+#include<iostream>
+#include <bits/stdc++.h>
+#include<stdlib.h>
+#include<time.h>
+using namespace std;
+int partition(int *a,int start,int end)
+{
+    int pindex=0;
+    int pivot=a[end];
+    for(int i=0;i<=end-1;i++)
+    {
+        if(a[i]<=pivot)
+        {
+            swap(a[i],a[pindex]);
+            pindex=pindex+1;
+        }
+    }
+    swap(a[pindex],a[end]);
+    return pindex;
+}
+int random_partition(int *a,int start,int end)  #it will choose the random element as pivot element
+{
+    int r;
+    srand(time(NULL));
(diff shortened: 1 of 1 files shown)


randomised_quicksort.cpp
```diff
@@ -0,0 +1,59 @@
+#include<iostream>
+#include <bits/stdc++.h>
```
ashwek: please do not use **bits/stdc++.h** as mentioned in the [README.md](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/README.md)

_as **bits/stdc++.h** includes all the standard header files it increases the compiler time & only works on GCC compilers_

randomised_quicksort.cpp
```diff
@@ -0,0 +1,59 @@
...
+        if(a[i]<=pivot)
+        {
+            swap(a[i],a[pindex]);
+            pindex=pindex+1;
+        }
+    }
+    swap(a[pindex],a[end]);
+    return pindex;
+}
+int random_partition(int *a,int start,int end)  #it will choose the random element as pivot element
```
ashwek: `#` is used for single line comments in **Python**. please replace `#` with `\\`

randomised_quicksort.cpp
```diff
@@ -0,0 +1,59 @@
...
+        int pindex=random_partition(a,start,end);
+        quicksort(a,start,pindex-1);
+        quicksort(a,pindex+1,end);
+    }
+}
+
+int main()
+{
+    int i;
+    int a[50000];
```
ashwek: instead of statically creating an array of 50,000 elements, use the size input (`n`) to create an array of required size.

```cpp
cin>>n;
int a[n];
```

statically creating an array of 50,000 elements wastes lot of memory.

ashwek:
please move your code to the __*Sorting*__ directory.

mouar:
ok i will do another commit after correction....

PR #301, opened 2019-09-30
 PigeonHole_Sorting.cpp

deepshi4156:
Pigeonhole sorting is a sorting algorithm that is suitable for sorting lists of elements where the number of elements and the number of possible key values are approximately the same.
It requires O(n + Range) time where n is number of elements in input array and ‘Range’ is number of possible values in array.


rajat77460:
Mam, the complexity of your code is O(range*n).
Please correct me if i am wrong.

deepshi4156:
No the complexity is O(range + n)

rajat77460:
Yes Mam, you are right.
https://www.geeksforgeeks.org/pigeonhole-sort/

PR #410, opened 2019-10-04
Add binary exponentiaion algorithm.

D4nnyLee:
Adding binary exponentiation algorithm.

### DIFF
--- a/Math/Binary_Exponentiation/binaryExponentiation.cpp
+++ b/Math/Binary_Exponentiation/binaryExponentiation.cpp
@@ -12,8 +12,9 @@ long long int solve(long long int b, long long int p, long long int m) {
 }
 int main(int argc, char const *argv[]) {
     // Compute the result of (b^p) % m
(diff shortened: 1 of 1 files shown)


Math/Binary_Exponentiation/binaryExponentiation.cpp
```diff
@@ -0,0 +1,19 @@
+#include <iostream>
+using namespace std;
+long long int solve(long long int b, long long int p, long long int m) {
```
madhur4127: don't you think the function name should be more descriptive, something which relates to binary exponentiation?

Math/Binary_Exponentiation/binaryExponentiation.cpp
```diff
@@ -0,0 +1,19 @@
+#include <iostream>
```
madhur4127: do you think a comment about what the code is doing will be better along with time/space complexity?

PR #425, opened 2019-10-05
Create sol1.cpp

manav014:
provided the solution for project euler question 26


manav014:
just tried to help out with project euler problem 26 bute force algorithm.

manav014:
just provided the in depth solution of 3 project euler problems and would be contributing other problems too

PR #434, opened 2019-10-06
create Euler's totient function

Aakash1822:
Create code for Euler's Totient function

Euler Totient function is a mathematical approach for an input n is a count of numbers(1,2,3,....,n) that are relatively prime to n i.e. the number whose GCD(Greatest Common Divisor) with n is 1.

Example:- It helps to find sequence next value like { 1,1,2,2,4,2,6,...........,10,4,......,n}

### DIFF
--- a/Data Structure/SegmentTree.cpp
+++ b/Data Structure/SegmentTree.cpp
@@ -1,0 +1,44 @@
+#include<iostream>
+using namespace std;
+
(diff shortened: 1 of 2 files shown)


Aakash1822:
I like to add this code which helps to solve eulers totient function and the related problems like sequence problem and same use-case.

Aakash1822:
#434

PR #484, opened 2019-10-10
update to dynamic sorting using pointer

akshay-gupta123:
use dynamic allocation to take as many as user wants and to prevent memory leak by using free
output of array in both increasing as well as decreasing order


akshay-gupta123:
i have added a sudoku algorithm which checks whether entered sudoku is valid or not

akshay-gupta123:
also added few other algorithm take a look at these as they are not present already

PR #519, opened 2019-10-13
Update prime factor finder to O(sqrt(n)) time complexity

Baranixowy:
Searching fuction is working now correctly.


Baranixowy:
Update prime factor finder.

PR #558, opened 2019-10-20
armstrong number.

haru-02:
this is a program to check if a given number is an Armstrong number or not .


theajaysingh:
This is not a armstrong number. it will work only for 3 digit not all numbers.

PR #566, opened 2019-10-21
Add ReverseTheNumber algorithm

TomWia9:
A simple algorithm that reverse the entered number

### DIFF
--- a/Others/ReverseTheNumber.cpp
+++ b/Others/ReverseTheNumber.cpp
@@ -1,0 +1,25 @@
+//This program reverse the number that you entered
+#include <iostream>
+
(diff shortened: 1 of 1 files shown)


Others/ReverseTheNumber.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+        return newNumber;
+}
+
+int main(){
+
+    int number;
+	std::cout << "Enter a number:";
+	std::cin >> number;
+    std::cout << "Reversed number is: " << reverse(number);
```
bhaumikmistry: check for char anything other than numbers?

PR #568, opened 2019-10-22
Added 3 way quick sort

saketk2019:


bhaumikmistry:
Please follow contribution guide.

```
Don't use bits/stdc++.h because this is quite Linux specific and slows down the compiler process.
```

Sorting/3_way_quick_sort.cpp
```diff
@@ -0,0 +1,93 @@
+#include<bits/stdc++.h>
```
bhaumikmistry: Don't use bits/stdc++.h because this is quite Linux specific and slows down the compiler process.

saketk2019:
Changes done

PR #574, opened 2019-10-23
Bug fix remove garbage entry, fix loop

MrAsimZahid:
Loop was taking storing garbage value. Fixed the loop now the algorithm is awesome.


MrAsimZahid:
Good to go.

PR #585, opened 2019-10-25
Prime

theajaysingh:
Here we can check prime upto 10^8 in O(1). It is very useful in Competitive programming.


cclauss:
If you are OK with these changes then I will merge this pull request.

PR #589, opened 2019-10-25
Matrix Exponentiation

kushagra98:
Matrix Exponentiation (also known as matrix power, repeated squaring) is a technique used to solve linear recurrences. This technique is very useful in competitive programming when dealing with linear recurrences (appears along Dynamic Programming).

Example to calculate the 10^18th fibonacci series term, it can not be done using Recursion, or DP but using matrix expo.


cclauss:
https://github.com/TheAlgorithms/C-Plus-Plus/pull/589/checks?check_run_id=334456290#step:9:46 Our automated tests are failing.

kushagra98:
There was a slight mistake in the code. It is fixed now.

cclauss:
Click __Details__ next to the ❌ below to see why tests are failing.

cclauss:
I am learning about cpplint and clang-format with your file.  I hope it is not too disturbing.

I am down to the last two issues but __I need your help__ to resolve them:
```
others/matrix_exponentiation.cpp:27:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
others/matrix_exponentiation.cpp:28:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
```

others/matrix_exponentiation.cpp
```diff
@@ -13,7 +13,7 @@ matrix.
 
 Steps for Matrix Expo
 1. Create vector F1 : which is the copy of B.
-2. Create transpose matrix (Learn more abput it on the internet)
+2. Create transpose matrix (Learn more abput it on the llernet)
```
cclauss: about and Internet are both misspelled.

others/matrix_exponentiation.cpp
```diff
@@ -1,5 +1,5 @@
 /*
-Matrix Exponentiation. If the problem can be solved with DP but constaints are
+Matrix Exponentiation. If the problem can be solved with DP but constalls are
```
cclauss: canstalls

kushagra98:
I don't why there are errors of spaces before and after '<<' and '>>'. I came across this problem for the first time. I think it is due to the compiler, maybe github uses a 64 bit compiler.

others/matrix_exponentiation.cpp
cclauss: Constraints?

kushagra98:
I have no clue how this compiler of github works. I'm still reading about it and trying to fix the code.

others/matrix_exponentiation.cpp
kushagra98: Yeah. I replaced all the 'int' with 'll'. I'll get this thing.

kushagra98:
Done!

cclauss:
cpplint is not a compiler.  It is a Python program that Google wrote to drive consistency in its code reviews.

kushagra98:
Thanks for the knowledge. I really had no idea about it. Actually I'm more into competitive programming, where you have to code fast. So generally we don't care about the spaces and tabs.
Now the code is good to go. Thanks for the contributions and your help as well. Much appreciated sir. 👍

PR #594, opened 2019-10-26
Implement 2. Add two numbers problem

ungps:


sarimurrab:
// TWO NUMBER ADDITION PROBLEM
#include<iostream>
using namespace std;
int main()
{
    int a,b;
    cin>>a>>b;
    cout<<a+b;
    return 0;
}

PR #599, opened 2019-10-28
Added josephus and large number addition

rehanvipin:
Added two new algorithms and implementations.
These use circular linked lists and mathematical calculations using strings


nikhilkala:
@rehanvipin 
From the code remove the comments which are not explanation.
The old code as comment makes the code less readable.

rehanvipin:
okay, will look into it ASAP

PR #603, opened 2019-10-28
Added an LCM Program in the math folder

jahnavibadam:


Math/LCM.txt
```diff
@@ -0,0 +1,23 @@
+#include <iostream>
+using namespace std;
+int main()
+{
+    int n1, n2, max;
+    cout << "Enter two numbers: ";
+    cin >> n1 >> n2;
```
AnupKumarPanwar: Kindly avoid taking the user input. Define your variables in the code. This makes the code easy to understand.

tjgurwara99:
Since the GCD algorithm for integers is already there, is there really a need for LCM? 
For example:
For any a,b integers, let g = gcd(a,b) and l = lcm(a,b) then it is easy to prove that g*l = a*b

PR #609, opened 2019-10-29
added Euler totient function

aniakubik:


nikhilkala:
@aniakubik 
the directory math should be used instead of Math

PR #610, opened 2019-10-30
Added new algo

kapilk535:


ashwek:
@kapilk535 thank you for contributing, but we already have an implementation of Decimal to Binary.
Closing PR.

PR #617, opened 2019-10-31
Create suffix tree

BLANK19999:
Algorithm name-Suffix tree for serching some pattern  in a string
this is suffix tree data structure implementation in c++ for serching some
pattern inside string

### DIFF
--- a/strings/suffix_tree.cpp
+++ b/strings/suffix_tree.cpp
@@ -1,14 +1,5 @@
 /*
-
-
-Algorithm name-Suffix tree for serching some pattern  in a string
-this is suffix tree data structure implementation in c++ for serching some
-pattern inside string
-
-*/
-
-/*
-HACKTOBERFEST
+Suffix tree facilitates searching for patterns in a string
 */
 
 #include<bits/stdc++.h>
@@ -20,6 +11,7 @@ struct suffix//creating structure name suffix
 	int end=0;//0 means not a leaf node  and 1 means it is a leaf node
 	
 };
+
 struct suffix * BuildSuffixTree(struct suffix * root,string s, int start, int size){
 	if(start==size&&root==NULL){//cheking if this is last char of string and that node does not exist already
 		struct suffix *temp;
@@ -27,8 +19,8 @@ struct suffix * BuildSuffixTree(struct suffix * root,string s, int start, int si
 		temp->end=1;
 		root=temp;
 	}
-	else if(start==size){//cheking if this is last char of string but node exist already 
-
+	else if(start==size){
+		//cheking if this is last char of string but node exist already 
 	}
 	else if(root!=NULL){//cheking for this is not last char in string and node for that char does not exist
 		//cout<<s;
@@ -44,10 +36,9 @@ struct suffix * BuildSuffixTree(struct suffix * root,string s, int start, int si
 		root->m[s[start]-'a']=BuildSuffixTree( root->m[s[start]-'a'], s, start+1, size );
 
 	}
-	
-	
 	return root;
 }
+
 void checkPatter(struct suffix *root, string pattern,int start ,int size){
 	if(start==size){// if the start pointer has reached to the end 
 		cout<<"YES this pattern exist in the string ";
@@ -61,6 +52,7 @@ void checkPatter(struct suffix *root, string pattern,int start ,int size){
 	}
 
 }
+
 int main(){
 	struct suffix *root = NULL;
 
@@ -80,7 +72,4 @@ int main(){
 	string pattern;
 	cin>>pattern;
 	checkPatter(root,pattern,0,pattern.length());
-
-
-
 }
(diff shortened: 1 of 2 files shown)


strings/suffix_tree.cpp
```diff
@@ -0,0 +1,75 @@
+/*
+Suffix tree facilitates searching for patterns in a string
+*/
+
+#include<bits/stdc++.h>
```
Panquesito7, line 5: Please don't use `bits/stdc++.h` as this is quite Linux specific and slows down the compilation process.
Also, in order to use `std::cout` or `std::cin`, you **must** add the `iostream` library.

strings/suffix_tree.cpp
```diff
@@ -0,0 +1,75 @@
...
+using namespace std;
```
Panquesito7, line 6: Don't use namespace using-directives.
Use using-declarations instead.

strings/suffix_tree.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+struct suffix//creating structure name suffix
+{
+	struct suffix*  m[26];//creating mapping for the 26 small latters alphabet
+	int end=0;//0 means not a leaf node  and 1 means it is a leaf node
+	
+};
+
+struct suffix * BuildSuffixTree(struct suffix * root,string s, int start, int size){
```
Panquesito7, line 15: Function not documented and missing space after function definition.

strings/suffix_tree.cpp
```diff
@@ -0,0 +1,75 @@
...
+	if(start==size&&root==NULL){//cheking if this is last char of string and that node does not exist already
+		struct suffix *temp;
+		temp=(struct suffix*)malloc(sizeof(struct suffix));//dynamically allocating memory
+		temp->end=1;
+		root=temp;
+	}
+	else if(start==size){
+		//cheking if this is last char of string but node exist already 
+	}
+	else if(root!=NULL){//cheking for this is not last char in string and node for that char does not exist
```
Panquesito7, line 25: Please use brackets in conditionals like:

```c++
if (var1 == var2) {
    // something here
} else {
    // something here
}
```
Same in other parts of the code.

strings/suffix_tree.cpp
```diff
@@ -0,0 +1,75 @@
...
+												//char left in pattern string to be searched
+		checkPatter(root->m[pattern[start]-'a'],pattern,start+1,size);//recursivly calling there child node
+	}
+	else{//start pointer has not reached to the end and there is some charecter missing in string 
+		cout<<"SORRY, this pattern does not exist in the string";
+	}
+
+}
+
+int main(){
```
Panquesito7, line 56: Please add a comment structure for the `main()` function:

```c++
/**
 * Main function
 */
```

strings/suffix_tree.cpp
```diff
@@ -0,0 +1,75 @@
...
+	cin>>s;
+
+	int n=s.length();
+	for(int i=0;i<n;i++)
+	root=BuildSuffixTree(root,s,i,n);
+
+	cout<<endl<<"enter the pattern you want to search :";
+	string pattern;
+	cin>>pattern;
+	checkPatter(root,pattern,0,pattern.length());
```
Panquesito7, line 74: Preferably, add `return 0;` at the end of the program.
It is a good programming practice.

strings/suffix_tree.cpp
```diff
@@ -0,0 +1,75 @@
...
+	cout<<"Enter your  string :";
+	cin>>s;
+
+	int n=s.length();
+	for(int i=0;i<n;i++)
+	root=BuildSuffixTree(root,s,i,n);
+
+	cout<<endl<<"enter the pattern you want to search :";
+	string pattern;
+	cin>>pattern;
```
Panquesito7, line 73: Why use `std::cin`? If the user types the name pattern with spaces
it will only save the phrase **before** the space.

For example, if the user types `Big Tree`, `std::cin` will only save `Big`, which is incorrect.
A way to solve this is to use `getline()` (which requires adding the `string` library), or change the variable type to `char[]`, and use `cin.getline()`.

strings/suffix_tree.cpp
```diff
@@ -0,0 +1,75 @@
...
+int main(){
+	struct suffix *root = NULL;
+
+	//dynamcally alocating the first root node
+	root=(struct suffix*)malloc(sizeof(struct suffix));
+	root->end=1;
+
+	string s;
+	cout<<"Enter your  string :";
+	cin>>s;
```
Panquesito7, line 65: Same here as above.

strings/suffix_tree.cpp
```diff
@@ -0,0 +1,75 @@
+/*
+Suffix tree facilitates searching for patterns in a string
```
Panquesito7, line 2: Please add the copyright notice like in #807.

PR #626, opened 2019-11-05
Created PigeonHole sorting algorithm

pooja795:


nikhilkala:
@pooja795 
This belongs in the sorting directory.
Push it into that directory.

PR #683, opened 2019-12-19
Repair warning: ISO C + + forbids converting a string constant to 'char *' [- wwrite strings]

Arctic2333:
Repair warning: ISO C + + forbids converting a string constant to 'char *' [- wwrite strings]


data_structure/TrieTree.cpp
```diff
@@ -1,97 +1,89 @@
...
+    nn -> isEndofWord = false;
     return nn;
 }
 
 // insert string into the trie
-void insert(trie *root, char *str)
-{
-    for (int i = 0; i < strlen(str); i++)
-    {
+void insert(trie * root, char * str) {
```
cclauss: Would it make sense to use __string__ instead of __char *__ in this instance?

data_structure/TrieTree.cpp
```diff
@@ -1,97 +1,89 @@
...
+    root -> isEndofWord = true;
 }
 
 // search a string exists inside the trie
-bool search(trie *root, char *str, int index)
-{
-    if (index == strlen(str))
-    {
-        if (!root->isEndofWord)
+bool search(trie * root, char * str, int index) {
```
cclauss: Would it make sense to use __string__ instead of __char *__ in this instance?

data_structure/TrieTree.cpp
```diff
@@ -1,97 +1,89 @@
...
 }
 /* removes the string if it is not a prefix of any  other 
  string, if it is then just sets the endofword to false, else 
  removes the given string*/
-bool deleteString(trie *root, char *str, int index)
-{
-    if (index == strlen(str))
-    {
-        if (!root->isEndofWord)
+bool deleteString(trie * root, char * str, int index) {
```
cclauss: Would it make sense to use __string__ instead of __char *__ in this instance?

data_structure/TrieTree.cpp
```diff
@@ -1,97 +1,89 @@
...
-    insert(root, "hello");
-    insert(root, "world");
-    int a = search(root, "hello", 0);
-    int b = search(root, "word", 0);
+int main() {
+    trie * root = createNode();
+    insert(root, const_cast < char * > ("hello"));
+    insert(root, const_cast < char * > ("world"));
+    int a = search(root, const_cast < char * > ("hello"), 0);
+    int b = search(root, const_cast < char * > ("word"), 0);
```
cclauss: Would it make sense to use __string__ instead of __char *__ in these four instances?

Arctic2333: I think it's better to use **string**. But I didn't edit this algorithm. I just added const_cast < char * > to repair **warning: ISO C + + forbids converting a string constant to 'char *' [- wwrite strings]**  Perhaps the best way is to rewrite the algorithm with string.

cclauss:
My sense with these in the future we should __#include <string>__ instead of __#include <string.h>__ as the C++ syntax is easier to understand and bring more options.

PR #687, opened 2019-12-20
#include<iosteream.h> --> <iostream>

Arctic2333:
#include<iosteream.h> --> <iostream>

The clrscr() function is a very old function with TC compiler. Because it does not conform to the standard of ANSI C, the general compiler does not support this function.

Replace clrscr () with system ("CLS"); statement

Conio. H is not a header file in the C standard library. It is not defined in the C standard library, ISO C and POSIX standards.
computer_oriented_statistical_methods : Most CPPs in this folder are written in Turbo C


computer_oriented_statistical_methods/false-position.cpp
```diff
@@ -0,0 +1,35 @@
+#include<stdlib.h>
+#include <math.h>
+#include <iostream>
+float eq(float i) {
+    return (pow(i, 3) - (4 * i) - 9);  // origial equation
+}
+int main() {
+    float a, b, z, c, m, n;
+    system("cls");
```
cclauss: __system("cls");__ will raise __-bash: cls: command not found__ on Linux and macOS.  This must be __system("clear");__ on Linux and macOS.

cclauss: Does __clear__ work on Windows?

PR #691, opened 2019-12-25
Add fast power

ibahadiraltun:
#### Description of Change
Implemented algorithm that computes a^b in O(logN) time recursively or linear. 

#### Checklist

### DIFF
--- a/math/fast_power.cpp
+++ b/math/fast_power.cpp
@@ -62,17 +62,17 @@ int main() {
         assert(fast_power_recursive(a, b) == std::pow(a, b));
         assert(fast_power_linear(a, b) == std::pow(a, b));
 
-        std::cout << "------ " << a << "^" << b << " = "<< 
+        std::cout << "------ " << a << "^" << b << " = "<<
             fast_power_recursive(a, b) << std::endl;
     }
 
     int64_t a, b;
     std::cin >> a >> b;
 
-    std::cout << a << "^" << b << " = "<< 
+    std::cout << a << "^" << b << " = "<<
         fast_power_recursive(a, b) << std::endl;
 
-    std::cout << a << "^" << b << " = "<< 
+    std::cout << a << "^" << b << " = "<<
         fast_power_linear(a, b) << std::endl;
 
     return 0;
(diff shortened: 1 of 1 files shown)


ibahadiraltun:
Fast Power algorithm that calculates a^b in O(logN) time.

ibahadiraltun:
Added sample tests

math/fast_power.cpp
```diff
@@ -0,0 +1,55 @@
...
+    return result;
+}
+
+int main() {
+    std::ios_base::sync_with_stdio(false);
+
+    std::cout << "3^5 is = " << fast_power_recursive(3, 5) << std::endl;
+    std::cout << "6^4 is = " << fast_power_recursive(6, 4) << std::endl;
+    std::cout << "2^7 is = " << fast_power_recursive(2, 7) << std::endl;
+    std::cout << "10^3 is = " << fast_power_recursive(10, 3) << std::endl;
```
cclauss: Would it be possible to create some tests of the form:
```cpp
assert (fast_power_recursive(10, 3) == 1000);
```
that would raise an error if __fast_power_recursive()__ was modified incorrectly?

Should we test zero and negative values in our tests?

ibahadiraltun: It would be possible to create tests using **assert()**.
This algorithm works for negative base but not for negative power. 
We can use the fact that aˆ(-b) = 1/(aˆb) but I do not think it is often used.

cclauss: Perhaps a Python example would help...
```

>>> 10 ** 3
1000
>>> -10 ** 3
-1000
>>> 10 ** -3
0.001
>>> -10 ** -3
-0.001
```
When run with each of these values, does the proposed algorithm return similar results?  If not, how does it fail?  Does if produce an invalid result or does it raise a helpful error?

cclauss: Perhaps tests of the form:
```cpp
assert (fast_power_recursive(10, 3) ==  std::pow(10, 3))
```

ibahadiraltun: Updated test form.

ibahadiraltun:
Added more random tests in appropriate form and provide a return value for negative powers.

cclauss:
Awesome contribution!  Thanks.

ibahadiraltun:
> Awesome contribution! Thanks.

Thank you!

PR #698, opened 2020-01-03
variable x is unused

ericcurtin:


sorting/heap_sort.cpp
```diff
@@ -1,5 +1,7 @@
 #include <iostream>
 
+using std::endl;
```
ericcurtin: This fixes the linter

PR #712, opened 2020-03-11
Update heap_sort.cpp

haowei790818:
The recursion root node in function 'heapify' should be the swapped one instead of length.

#### Checklist

Notes:


nikhilkala:
The present implementation seems correct to me and works correctly.
What is the need for the change?
could you elaborate on the description?
@haowei790818

NuuNaNa:
Hello

PR #718, opened 2020-04-14
Added number_of_positive_divisors.cpp in math directory

mann2108:
#### Description of Change

This algorithm use the prime factorization approach.
Any number can be written in multiplication of its prime factors.
Let N = P1^E1 * P2^E2 ... Pk^Ek
Therefore. number-of-divisors(N) = (E1+1) * (E2+1) ... (Ek+1).
Where P1, P2 ... Pk are prime factors and E1, E2 ... Ek are exponents respectively.

The Complexity of this algorithm is O(N).
If we have multiple queries then its complexity reduces to O(Q * sqrt(n)) using the precomputation of all primes with Sieve Eratosthenes. 

Example:- N=36
36 = (2^2 * 3^2)
number-of-divisors(36) = (2+1) * (2+1) = 9.
list of divisors of 36 = 1, 2, 3, 4, 6, 9, 12, 18, 36.

#### Checklist

- [X] Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTION.md#New-File-Name-guidelines)
- [X] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTION.md#Commit-Guidelines)
- [ ] Sort by alphabetical order

Notes:


mann2108:
@cclauss I have successfully open a PR. You can now review it. :)

mann2108:
Well for this I want your suggestion.
There are two ways either the algorithm work for only positive number or we can also handle negative numbers too.
Like take an example of 4 then there are total 6 divisors including negative numbers. 
4 = 1, 2, 4, -1, -2, -4.
Hence the only modification I guess its 2 times the current output.
But generally when we talking about the divisors we only consider positive values. So should I modify algorithm in such a way that it also count negative divisors as well as take negative inputs or just work around positive numbers. what you say ?

cclauss:
Let’s keep the answer as it already is for n = 4.
What do we want the answer to be if n = -4?

mann2108:
> Let’s keep the answer as it already is for n = 4.
> What do we want the answer to be if n = -4?
Actually my opinion is that if we are considering negative divisors for any number then we should consider for both positive as well negative input or none ?.
If we take -4 output as 6 then +4 output must be 6 as -4, -2 and -1 perfectly divides +4 too.

mann2108:
We can consider only positive divisors for any input -4 and +4. I just found one blog regarding this on wikipedia. 

Capture2

cclauss:
I like that approach.

math/number_of_positive_divisors.cpp
```diff
@@ -12,11 +12,11 @@
  *
  * Example:- N=36
  * 36 = (3^2 * 2^2)
- * number-of-divisors(36) = (2+1) * (2+1) = 9.
- * list of divisors of 36 = 1, 2, 3, 4, 6, 9, 12, 18, 36.
+ * number_of_positive_divisors(36) = (2+1) * (2+1) = 9.
```
cclauss, line 15: Please also document the results of `number_of_positive_divisors(-36)` so that users know what to expect if they provide a negative number to your algorithm.

mann2108:
@cclauss updated with requested changes you can now review it.

math/number_of_positive_divisors.cpp
mann2108: Done !

mann2108:
@cclauss Thanks !

PR #724, opened 2020-04-22
fix: reverse check for empty list

mariangusatu:
I've fixed a small bug that occurs when the linked list is empty and we try to reverse it.


data_structure/linked_list.cpp
```diff
@@ -0,0 +1,137 @@
+#include <iostream>
```
nikhilkala, line 1: @mariangusatu 
Could you point out exact lines where code was changed as because of rename, code changes are not visible.

PR #771, opened 2020-05-21
add_string code added

Ravisheoran486:
i have added a code to add two strings with a space.


Ravisheoran486:
i have corrected it again

kvedala:
Thank you for the contribution. But this is not an algorithm.

PR #772, opened 2020-05-21
fix #765: Doubly Linked List algorithm bug

### LINKED ISSUE #765: [BUG]SegFault in remove(x) DLL
## Description

the remove() function throws a segmentation fault. 

## Expected Behavior

The function should remove the parsed element.

## Actual Behavior

The function fails at "t->prev->next = t->next;" and throws a seg fault

## Possible Fix

Still working on it, just wanted to report it... I don't know C++ well.

## Steps to Reproduce

1. insert some values
2. print if you want
3. delete something

## Context

I was trying to remove a node from the DLL

SamratDe:
Fixed the remove function which was causing segmentation fault while removing first and last elements in the doubly linked list.


SamratDe:
@ayaankhan98 please check now. I have handled all automated failing tests.

cclauss:
> use class instead of struct

Why is that required??

cclauss:
Is the user warned if the item that they ask to remove is not present?

ayaankhan98:
> > use class instead of struct
> 
> Why is that required??

i think using object oriented concepts code becomes more managable and eaisly understandable if we have a class doubly_linked_list and all the functions associated with this container wrapped within this class

SamratDe:
@ayaankhan98 I have wrapped all functions in a double_linked_list class

cclauss:
> using object oriented concepts code becomes more managable and eaisly understandable

There is a lot of pushback against OOP.  For example the Go language has completely abandoned classes.  So while classes might help, let’s not make them a requirement in all submission.

ayaankhan98:
> > using object oriented concepts code becomes more managable and eaisly understandable
> 
> There is a lot of pushback against OOP. For example the Go language has completely abandoned classes. So while classes might help, let’s not make them a requirement in all submission.

okay sure @cclauss

PR #784, opened 2020-05-23
Updated DLL with n_th_location adding feature & added tarjan algorithm for bridge finding

offamitkumar:
1. in the data structure section, the doubly-link-list is added a new function to add a number at the nth location.

2. in graph algorithm section, add Tarjan algorithm for bridge finding


cclauss:
One .cpp file per pull request please.

PR #789, opened 2020-05-24
feat : bridge finding with tarjan algorithm

### LINKED ISSUE #783: [FEATURE] bridge finding algorithm with Trajan Algorithm
In graph section, the bridge finding algorithm with Trajan algorithm is not present. 
I would like to implement it.

offamitkumar:
Fixes #783

Added Bridge finding with Tarjan's Algorithm.


cclauss:
If you make modifications on branches instead on master then you will not block yourself.

PR #802, opened 2020-05-27
Added documentation

ayaankhan98:


sorting/quick_sort.cpp
```diff
@@ -1,16 +1,33 @@
...
+ *      correct position in sorted array and put all smaller elements (samller
+ *      than x) before x, and put all greater elements (greater than x) after
+ *      x. All this should be done in linear time
+ *
  */
 
 #include <cstdlib>
 #include <iostream>
 
 int partition(int arr[], int low, int high) {
```
kvedala, line 29: can you add a brief comment to the individual functions as well.

kvedala:
👍 
the comment blocks should be like this
```c
/** <- start of comment block 


      end of comment block 
 */
```

ayaankhan98:
i think all seems to be fine now

kvedala:
yes, 👍

ayaankhan98:
finally merging this

sorting/quick_sort.cpp
```diff
@@ -37,14 +69,14 @@ void quickSort(int arr[], int low, int high) {
...
+// prints the array after sorting
 void show(int arr[], int size) {
     for (int i = 0; i < size; i++)
         std::cout << arr[i] << " ";
     std::cout << "\n";
 }
 
 // Driver program to test above functions
-
 int main() {
```
kvedala, line 80: @ayaankhan98 
There is a dynamic memory allocation for the array `int *arr` on line 84 which should be freed before the program termination.

Add the line just before the return of the main function.
```c
delete [] arr;
```

ayaankhan98: #803

PR #803, opened 2020-05-27
free dynamically allocated memory 

ayaankhan98:
added `delete [] arr;` to deallocate the dynamically allocated memory after task finishes


sorting/quick_sort.cpp
```diff
@@ -37,14 +69,14 @@ void quickSort(int arr[], int low, int high) {
     }
 }
 
+// prints the array after sorting
```
kvedala, line 72: comment format

sorting/quick_sort.cpp
```diff
@@ -1,16 +1,42 @@
...
+ *      the pivot element at its correct position in sorted  
+ *      array, and places all smaller (smaller than pivot)  
+ *      to left of pivot and all greater elements to right  
+ *      of pivot 
+ *
+ */
+
 int partition(int arr[], int low, int high) {
-    int pivot = arr[high];    // pivot
+    int pivot = arr[high];    // taking the last element as pivot
```
deadshotsb, line 39: Taking the extreme elements as pivot lands up the algorithm in the worst case scenario in case where data is sorted. So, if you choose your pivot as the middle element or try out a random element that will land up in the average case complexity.

kvedala:
@ayaankhan98 This code is still not working correctly.
I used this modification for the tests.

```cpp
#include <algorithm>
#include <cassert>

/** Driver program to test above functions */
int main() {
    for (int num_test = 0; num_test < 10; num_test++) {
        int size = std::rand() % 500;
        std::cout << "Iter# " << num_test << " Array size: " << size;

        int *arr = new int[size];

        for (int i = 0; i < size; ++i) {
            arr[i] = std::rand() % 500 - 250;
        }
        quickSort(arr, 0, size);
        assert(std::is_sorted(arr, arr + size));
        delete[] arr;
        std::cout << " Passed\n";
    }
    return 0;
}
```

ayaankhan98:
yeah sure @kvedala  i will check it out and create a PR in fix of this.

PR #807, opened 2020-05-29
feat: Modern trie code with better memory management

Anmol3299:
#### Description of Change
The older implementation has very big memory leaks as there is no delete operation for the raw pointers used and assigned. The newer implementation uses modern c++ code and solves the memory leak issue too. I have tried to document the code as well as I could have.

#### Checklist

Notes: Modernize code and fix memory leaks of trie data structure.

### DIFF
--- a/data_structure/trie_modern.cpp
+++ b/data_structure/trie_modern.cpp
@@ -1,0 +1,156 @@
+#include <iostream>
+#include <memory>  // for std::shared_ptr<>
+#include <string>  // for std::string class
+
+/**
+ * A basic implementation of trie class to store only lower-case strings.
+ * You can extend the implementation to all the ASCII characters by changing the
+ * value of @ ALPHABETS to 128.
+ */
+class Trie {
+ private:
+    static constexpr size_t ALPHABETS = 26;
+
+    /**
+     * Structure of trie node.
+     * This struct doesn't need a constructor as we are initializing using
+     * intializer list which is more efficient than if we had done so with
+     * constructor.
+     */
+    struct TrieNode {
+        // An array of pointers of size 26 which tells if a character of word is
+        // present or not.
+        std::shared_ptr<TrieNode> character[ALPHABETS]{nullptr};
+
+        bool isEndOfWord{false};
+    };
+
+    /**
+     * Function to check if a node has some children which can form words.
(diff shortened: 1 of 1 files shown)


data_structure/trie_modern.cpp
```diff
@@ -0,0 +1,156 @@
+#include <iostream>
```
kvedala, line 1: Please add the comment structure at the Beijing of the file. This enables the document generating tool to understand and read the file. refer -> https://github.com/TheAlgorithms/C-Plus-Plus/blob/16fec03ac899cdc847f20e553de7850a83caaa71/search/binary_search.h

data_structure/trie_modern.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include <memory>  // for std::shared_ptr<>
+#include <string>  // for std::string class
```
kvedala, line 3: the `<string>` header is specific to Microsoft Visual C++. To make this generic, please use the following:
```c
#ifdef _MSC_VER
#include <string>
#else
#include <cstring>
#endif
```

data_structure/trie_modern.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+            curr = curr->character[index];
+        }
+        curr->isEndOfWord = true;
+    }
+
+    /**
+     * Search if a word is present in trie or not.
+     * @param word which is needed to be searched in the trie.
+     * @return if the word is found in the trie and isEndOfWord is set to true,
```
kvedala, line 121: for return statements, you can also break then into two as such:
```
* @return True if word is found and ...
* @return False if word is not found and ...
```

data_structure/trie_modern.cpp
```diff
@@ -0,0 +1,156 @@
...
+    // Function to remove the word which calls the helper function.
+    void removeWord(const std::string& word) {
+        m_root = removeWordHelper(word, m_root, 0);
+    }
+
+ private:
+    std::shared_ptr<TrieNode> m_root;
+};
+
+int main() {
```
kvedala, line 148: For the main, you can use a simple comment block like:
```
/**
 * Main function
 */
```

Anmol3299: Pushed with all the suggestions

PR #816, opened 2020-06-02
feat: Add BFS and DFS algorithms to check for cycle in a directed graph

Anmol3299:
#### Description of Change
Add BFS and DFS algorithms to check for cycle in a directed graph

#### Checklist

Note: The non-const reference is needed in this case

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/816"></a>


Anmol3299:
> graph/cycle_check_directed_graph.cpp:186: Is this a non-const reference? If so, make const or use a pointer: std::vector& state
> **Use of non const reference parameter violates google coding style**
> [Avoid non-const reference parameters [duplicate]](https://stackoverflow.com/questions/31632716/avoid-non-const-reference-parameters)

I did some "googling" and it seems that it is only the google standard to pass pointers instead of non-const references. However, various other sources suggest that it is quite okay and in-fact a better way than the pointers.
https://www.quora.com/Is-it-a-bad-practice-to-pass-arguments-by-reference-in-C-since-it-obscures-at-call-site-which-arguments-will-be-modified-while-passing-pointer-makes-it-explicit

Even the isocpp which is the cpp standard, suggests using non-const references: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rf-conventional

kvedala:
[ISO CPP Core guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#fcall-parameter-passing) do mention the benefit of using constant references. Hence, google, after a thorough due-diligence, came up with the Google CPP skyline [guidelines](https://google.github.io/styleguide/cppguide.html#Reference_Arguments); which is what the `cpplint` and this repository follow.

Anmol3299:
> [ISO CPP Core guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#fcall-parameter-passing) do mention the benefit of using constant references. Hence, google, after a thorough due-diligence, came up with the Google CPP skyline [guidelines](https://google.github.io/styleguide/cppguide.html#Reference_Arguments); which is what the `cpplint` and this repository follow.

Like I said, the non-const reference is needed here because I'm modifying it within the function. Returning the value would lead to copying of that vector which would be a costly operation

kvedala:
> > [ISO CPP Core guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#fcall-parameter-passing) do mention the benefit of using constant references. Hence, google, after a thorough due-diligence, came up with the Google CPP skyline [guidelines](https://google.github.io/styleguide/cppguide.html#Reference_Arguments); which is what the `cpplint` and this repository follow.
> 
> Like I said, the non-const reference is needed here because I'm modifying it within the function. Returning the value would lead to copying of that vector which would be a costly operation

Then you'd have to use a pointer per the docs

Anmol3299:
Please use c++17 standard.

kvedala:
> Please use c++17 standard.

This is just a warning and can be ignored for now.
However, the error is regarding something else.
Screen Shot 2020-06-03 at 10 16 39 AM
Also, per the above discussed standards, the use of `&&` for function input variables is prone to bugs.

Anmol3299:
> > Please use c++17 standard.
> 
> This is just a warning and can be ignored for now.
> However, the error is regarding something else.
> Screen Shot 2020-06-03 at 10 16 39 AM
> Also, per the above discussed standards, the use of `&&` for function input variables is prone to bugs.

&& signifies a rvalue reference which is clearly not an error and is intended.

kvedala:
[see this](https://google.github.io/styleguide/cppguide.html#Rvalue_references) 
I believe the static analyzer is failing due to that.

Anmol3299:
@kvedala there's been no action on the pull request yet. Are there any issues?

graph/cycle_check_directed_graph.cpp
```diff
@@ -0,0 +1,302 @@
+/**
+ * Copyright 2020
+ * @file cycle_check_directed graph.cpp
+ *
+ * @brief BFS and DFS algorithms to check for cycle in a directed graph.
+ *
+ * @author Anmol3299
```
kvedala, line 7:
```suggestion
 * @author [Anmol3299](mailto:mittalanmol22@gmail.com)
```

graph/cycle_check_directed_graph.cpp
```diff
@@ -0,0 +1,302 @@
...
+ * contact: mittalanmol22@gmail.com
```
kvedala, line 8:
```suggestion
```

graph/cycle_check_directed_graph.cpp
```diff
@@ -0,0 +1,302 @@
...
+        }
+        m_adjList[source].emplace_back(destination);
+    }
+
+ private:
+    unsigned int m_vertices;
+    AdjList m_adjList;
+};
+
+class CycleCheck {
```
kvedala, line 153: can you please add documentation just like for the `class Graph`

kvedala:
@Anmol3299 any updates not this?

Anmol3299:
> @Anmol3299 any updates on this? 

Ahh. Sorry, I completely forgot about this. I'll push the update today.

Anmol3299:
I kinda don't understand what the error really is.

kvedala:
> I kinda don't understand what the error really is.

1.`std::remove_reference_t` is defined in `<type_traits>` which is not included

lets fix this first

Anmol3299:
> > I kinda don't understand what the error really is.
> 
> 1.`std::remove_reference_t` is defined in `<type_traits>` which is not included
> 
> lets fix this first

type_traits is include in line 14 but for some reason it didn't work with remove_reference_t. So i changed to remove_reference<>::type which fixes the issue

kvedala:
Screen Shot 2020-07-13 at 9 13 17 AM

This implementation is compatible with C++17 and above. The repo uses C++11 to ensure that the algorithms can be implemented on low-power microcontrollers as well. The error was not observed because the `graph` folder in not included in the CMake build system. Would be great to maintain consistency throughout. 
@Anmol3299 , would you like to help out in adding the graph folder to CMake build system of the repo? This may become an extensive work, depending on the errors we may come across.

Anmol3299:
> Screen Shot 2020-07-13 at 9 13 17 AM
> 
> This implementation is compatible with C++17 and above. The repo uses C++11 to ensure that the algorithms can be implemented on low-power microcontrollers as well. The error was not observed because the `graph` folder in not included in the CMake build system. Would be great to maintain consistency throughout.
> @Anmol3299 , would you like to help out in adding the graph folder to CMake build system of the repo? This may become an extensive work, depending on the errors we may come across.

I am currently busy preparing for my job interviews so I don't have the time for all of the graph folder.
However, I can quickly convert my code to be compatible with C++11

PR #820, opened 2020-06-02
used  boolean instead of int to avoid and optimized code using Sieving till root

iamrajiv:
#### Description of Change

used  boolean instead of int to avoid TLE for large N in `sieve_of_eratosthenes.cpp` since boolean takes 2 bytes whereas int takes 4 bytes and also optimized code using Sieving till root

#### Checklist

Notes:

### DIFF
--- a/math/sieve_of_eratosthenes.cpp
+++ b/math/sieve_of_eratosthenes.cpp
@@ -1,5 +1,5 @@
 /*
- * Sieve of Eratosthenes is an algorithm to find the primes 
+ * Sieve of Eratosthenes is an algorithm to find the primes
  * that is between 2 to N (as defined in main).
  *
  * Time Complexity  : O(N * log N)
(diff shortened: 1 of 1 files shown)


iamrajiv:
@mishraabhinn Should I make issue first about previous code because previous `sieve` code was running for `N` time in the loop but It can be optimized by running `sqrt(N)` which gives the same result. Also, the previous code was using `int` which was taking 4 bytes but can be improved by using boolean which takes 2 bytes.

mishraabhinn:
I got the point and agree on it. 
You are getting an error because **std::** is not used before cout, cin etc.  Do that. 
Don't use using namespace std. 
Indent your code accordingly , so it conforms to Google's coding style guides. 
You can use clang-format or do that manually.

kvedala:
as an addendum to @mishraabhinn's nice review, please note:
1. this repository is for educational purposes and not to provide 100% performance optimized code
2. hence, if you would like to submit a different implementation for an existing algorithm, please submit it as a new file in the corresponding folder.

iamrajiv:
thank you for the suggestion @kvedala @mishraabhinn I will add a different implementation for the existing algorithm and for this PR I will revert my code and will remove `bits/stdc++.h` from existing `sieve_of_eratosthenes.cpp` because it is importing `bits/stdc++.h`

iamrajiv:
thank you for the suggestion @kvedala @mishraabhinn I will add a different implementation for the existing algorithm and this PR I will revert my code and will also remove `using namespace std;` from existing `sieve_of_eratosthenes.cpp`  and add `std::` wherever it is needed

kvedala:
> thank you for the suggestion @kvedala @mishraabhinn I will add a different implementation for the existing algorithm and this PR I will revert my code and will also remove `using namespace std;` from existing `sieve_of_eratosthenes.cpp` and add `std::` wherever it is needed

Thank you @iamrajiv. You can simply close this PR and start afresh.

PR #821, opened 2020-06-03
added classical_sieve_of_eratosthenes.cpp

iamrajiv:
#### Description of Change

Added `classical_sieve_of_eratosthenes.cpp` in math

#### Checklist

Notes:

### DIFF
--- a/math/classical_sieve_of_eratosthenes.cpp
+++ b/math/classical_sieve_of_eratosthenes.cpp
@@ -8,13 +8,12 @@
 
 #include <iostream>
 
-bool* isprime = new bool[100000];
-
 /**
  * This is the function that finds the primes and eliminates
  * the multiples.
 **/
-void sieve(int N) {
+void sieve(int N, bool* isprime ) {
+
   isprime[0] = true;
   isprime[1] = true;
   for (int i = 2; i * i <= N; ++i) {
@@ -29,7 +28,7 @@ void sieve(int N) {
 /**
  * This function prints out the primes to STDOUT
 **/
-void print(int N) {
+void print(int N, bool* isprime ) {
   for (int i = 1; i <= N; i++) {
     if (!isprime[i]) {
       std::cout << i << ' ';
@@ -39,8 +38,10 @@ void print(int N) {
 }
 
 int main() {
+  fastio();
   int N = 100;
-  sieve(N);
-  print(N);
+  bool* isprime = new bool[N];
+  sieve(N, isprime);
+  print(N, isprime);
   delete[] isprime;
 }
--- a/math/classical_sieve_of_eratosthenes.cpp
+++ b/math/classical_sieve_of_eratosthenes.cpp
@@ -5,7 +5,7 @@
  * Space Complexity : O(N)
  */
 
-#include <iostream> // for io operations
+#include <iostream>  // for io operations
 
 /**
  * This is the function that finds the primes and eliminates
@@ -42,6 +42,6 @@ int main() {
   sieve(N, isprime);
   print(N, isprime);
   delete[] isprime;
-  
+
   return 0;
 }
--- a/math/classical_sieve_of_eratosthenes.cpp
+++ b/math/classical_sieve_of_eratosthenes.cpp
@@ -1,17 +1,14 @@
-/**
- * Classical Sieve of Eratosthenes is an algorithm to find the primes
+/** Classical Sieve of Eratosthenes is an algorithm to find the primes
  * that is between 2 to N (as defined in main) in fastest way.
- *
  * Time Complexity  : O(N * log(log(N)))
  * Space Complexity : O(N)
-**/
+ */
 
 #include <iostream>
 
-/**
- * This is the function that finds the primes and eliminates
+/** This is the function that finds the primes and eliminates
  * the multiples.
-**/
+ */
 void sieve(int N, bool* isprime ) {
   isprime[0] = true;
   isprime[1] = true;
@@ -24,9 +21,8 @@ void sieve(int N, bool* isprime ) {
   }
 }
 
-/**
- * This function prints out the primes to STDOUT
-**/
+/** This function prints out the primes to STDOUT
+ */
 void print(int N, bool* isprime ) {
   for (int i = 1; i <= N; i++) {
     if (!isprime[i]) {
--- a/math/classical_sieve_of_eratosthenes.cpp
+++ b/math/classical_sieve_of_eratosthenes.cpp
@@ -38,7 +38,7 @@ void print(int N, bool* isprime ) {
 
 int main() {
   int N = 100;
-  bool* isprime = new bool[N];
+  bool *isprime = new bool[N];
   sieve(N, isprime);
   print(N, isprime);
   delete[] isprime;
--- a/math/classical_sieve_of_eratosthenes.cpp
+++ b/math/classical_sieve_of_eratosthenes.cpp
@@ -13,7 +13,6 @@
  * the multiples.
 **/
 void sieve(int N, bool* isprime ) {
-
   isprime[0] = true;
   isprime[1] = true;
   for (int i = 2; i * i <= N; ++i) {
--- a/math/classical_sieve_of_eratosthenes.cpp
+++ b/math/classical_sieve_of_eratosthenes.cpp
@@ -7,6 +7,7 @@
 **/
 
 #include <iostream>
+#include <climits>
 
 bool isprime[INT_MAX];
 
--- a/math/classical_sieve_of_eratosthenes.cpp
+++ b/math/classical_sieve_of_eratosthenes.cpp
@@ -37,7 +37,6 @@ void print(int N, bool* isprime ) {
 }
 
 int main() {
-  fastio();
   int N = 100;
   bool* isprime = new bool[N];
   sieve(N, isprime);
--- a/math/classical_sieve_of_eratosthenes.cpp
+++ b/math/classical_sieve_of_eratosthenes.cpp
@@ -1,0 +1,48 @@
+/*
+ * Classical Sieve of Eratosthenes is an algorithm to find the primes
+ * that is between 2 to N (as defined in main) in fastest way.
+ *
+ * Time Complexity  : O(N * log(log(N)))
+ * Space Complexity : O(N)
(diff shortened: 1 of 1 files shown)


math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Time Complexity  : O(N * log(log(N)))
+ * Space Complexity : O(N)
+ */
+
+#include <iostream>
+
+#define MAX 10000000
+
+bool isprime[MAX];
```
ayaankhan98, line 13: use dynamic memory allocation instead of using this much large array

math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+/*
```
ayaankhan98, line 15: please change commennt style 
it should be like
```
/**
 *
 *
 */
```

math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -1,21 +1,19 @@
...
  *
  * Time Complexity  : O(N * log(log(N)))
  * Space Complexity : O(N)
- */
+**/
 
 #include <iostream>
 
-#define MAX 10000000
+bool isprime[INT_MAX];
```
kvedala, line 12: this is 1GB of RAM!!!
`INT_MAX = 2147483647 bytes / (1024 * 1024 * 1024) =2GB`

math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -1,21 +1,20 @@
...
  * Time Complexity  : O(N * log(log(N)))
  * Space Complexity : O(N)
- */
+**/
 
 #include <iostream>
+#include <climits>
 
-#define MAX 10000000
+bool isprime[INT_MAX];
```
ayaankhan98, line 12: use dynamic memory allocation
try to allocate the memory using the `new` keyword instead of using `INT_MAX`
and also deallocate the memory after use using `delete` keyword

if you wish to allocate a bool array of size 20 (say) then you should write
`bool *arr = new bool[20];`
and to delete this allocated memory write
`delete [] arr;`

i hope now things are clear

kvedala: To add to this, `INT_MAX` bool array will try to allocate **2GB** of ram!! use only as much needed

iamrajiv: @kvedala can you help me how to implement this dynamically?

math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -0,0 +1,46 @@
...
+ * Classical Sieve of Eratosthenes is an algorithm to find the primes
+ * that is between 2 to N (as defined in main) in fastest way.
+ *
+ * Time Complexity  : O(N * log(log(N)))
+ * Space Complexity : O(N)
+**/
+
+#include <iostream>
+
+bool* isprime = new bool[100000];
```
ayaankhan98, line 11: still you are allocating a large amount of memory, let me explain you a bit more clearly

suppose you need to find the prime numbers between 2 and 30, in this case N = 30
so it is sufficient to have an array of size 30 instead of having an array of size 1000000, because
we will use only the first 30 blocks of memory rest all the blocks are useless in this case then there is no need to
take those resources from the system which are not required.
so what we can do in this situation is this we can allocate the memory of 30 blocks by writing
`bool *isPrime = new bool[30];`

and in this particular case you are taking the value of N in the `main()` function so declare this
array of bool in `main()` after taking the value of N, and in order to pass this array of bools to some other functions just pass the pointer to this array by reference to that function, now you don't need to worry to much about how to pass pointer to reference because in C/C++ arrays are automatically passed by reference.

and one last thing is after the program finished then you have to deallocate the resources which you have taken from the system thus it's good to deallocate this memory which you have taken earlier, and to delete this memory write
`delete [] isPrime;`

now it's a  bit more clear let me know if you have any other doubt.

math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -39,8 +38,10 @@ void print(int N) {
 }
 
 int main() {
+  fastio();
```
ayaankhan98: syntax error `fastio()` is not defined

math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -39,8 +38,10 @@ void print(int N) {
...
   int N = 100;
-  sieve(N);
-  print(N);
+  bool* isprime = new bool[N];
```
ayaankhan98, line 41: let it be `bool *isPrime = new bool[N];` or `bool *is_prime = new bool[N];`

math/classical_sieve_of_eratosthenes.cpp
iamrajiv: @ayaankhan98 I have already removed it in the last commit

math/classical_sieve_of_eratosthenes.cpp
iamrajiv: @ayaankhan98 I think name `isprime` looks good because in other `sieve_of_eratosthenes.cpp` file it has also the same name `isprime`.

iamrajiv: @ayaankhan98 I think name `isprime` looks good because in other `sieve_of_eratosthenes.cpp` file it has the same name `isprime` so it maintains the consistency.

ayaankhan98: if the algorithm is already present, then i think there is no need of the same stuff what you can do is that you can improve the old file if you see some mistake or you can optimize the previously written algorithms.

iamrajiv: @ayaankhan98 kindly check discussion on #820

ayaankhan98: @iamrajiv oh okay so this is the optimised version of algorithm which is already present, that's what i was suggesting also to optimize the already present algorithm

iamrajiv: @ayaankhan98 Initially I was doing the same i.e. optimising the existing algorithms but @kvedala and @mishraabhinn suggested to submit a different implementation for an existing algorithm and I agree with them because this repository is for educational purposes and I think people should know different approaches to solve the same problem and that what is has been done in `sieve_of_eratosthenes` since the previous code was not been implemented using Sieving till root but this code uses Sieving till root approach so it makes people learn a new approach.

math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -0,0 +1,45 @@
+/**
```
ayaankhan98, line 1:
```
/** Brief description which ends at this dot. Details follow
 *  here.
 */
```
please use this doxygen comment format

math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -0,0 +1,45 @@
...
+ * Classical Sieve of Eratosthenes is an algorithm to find the primes
+ * that is between 2 to N (as defined in main) in fastest way.
+ *
+ * Time Complexity  : O(N * log(log(N)))
+ * Space Complexity : O(N)
+**/
+
+#include <iostream>
+
+/**
```
ayaankhan98, line 11:
```
/** Brief description which ends at this dot. Details follow
 *  here.
 */
```
please use this doxygen comment format

math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -0,0 +1,45 @@
...
+  for (int i = 2; i * i <= N; ++i) {
+    if (!isprime[i]) {
+      for (int j = i * i; j <= N; j = j + i) {
+        isprime[j] = true;
+      }
+    }
+  }
+}
+
+/**
```
ayaankhan98, line 27:
```
/** Brief description which ends at this dot. Details follow
 *  here.
 */
```
please use this doxygen comment format

math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -0,0 +1,41 @@
...
+ * Time Complexity  : O(N * log(log(N)))
+ * Space Complexity : O(N)
+ */
+
+#include <iostream>
+
+/** This is the function that finds the primes and eliminates
+ * the multiples.
+ */
+void sieve(int N, bool* isprime ) {
```
ayaankhan98, line 12: change this be`void sieve(int N, bool *isprime) {`

math/classical_sieve_of_eratosthenes.cpp
```diff
@@ -0,0 +1,41 @@
...
+      for (int j = i * i; j <= N; j = j + i) {
+        isprime[j] = true;
+      }
+    }
+  }
+}
+
+/** This function prints out the primes to STDOUT
+ */
+void print(int N, bool* isprime ) {
```
ayaankhan98, line 26: change this to be`void print(int N, bool *isprime) {`

kvedala:
👍 Great effort by the author and reviewer. 
Could you please update documentation with standard similar to #807 
Thank you.

iamrajiv:
@ayaankhan98 please review my last commit

iamrajiv:
@kvedala @ayaankhan98 please review my last commit

iamrajiv:
@kvedala what about the space complexity since last code was using an int static array  but this code uses bool dynamic array

ayaankhan98:
> @kvedala what about the space complexity since last code was using an int static array but this code uses bool dynamic array

i think you can fix that too in that file

kvedala:
> @kvedala what about the space complexity since last code was using an int static array but this code uses bool dynamic array

try to do a print of `sizeof(bool)` and `sizeof(char)` -> they both return `1` i.e., both a boolean type and a character type occupy 1 byte of memory when compiled using GCC, MSVC. Some compilers like the Intel Compiler might be able to optimize it further to a 1 bit as one would not tally expect but this is not a necessity. 

That being said, to your point, to reduce the memory space, all it needs is a change from `int` to `bool` and the overall algorithm remains the same.

PR #825, opened 2020-06-04
feat: Add a calculator

Panquesito7:
#### Description of Change
Added a calculator in the `math` folder.

#### Checklist

Notes: Tested with Visual Studio 2019.

### DIFF
--- a/math/calculator.cpp
+++ b/math/calculator.cpp
@@ -21,13 +21,8 @@ struct Node {
  * Prototypes
  */
 
-void requestData();
+void requestData(Node*);
 void pause();
-
-Node* new_node = new Node();
-
-int opt = 0;
-float number;
 
 /**
  * --------------
@@ -36,7 +31,8 @@ float number;
  */
 
 int main() {
-    requestData();
+    Node* new_node = new Node();
+    requestData(new_node);
 
     pause();
     return 0;
@@ -57,7 +53,8 @@ void pause() {
  * Function to call the calculator.
  */
 
-void requestData() {
+void requestData(Node* new_node) {
+    int opt = 0; float number;
     int amount_to_sum = 0, amount_to_subtract = 0, amount_to_multiply = 0,
         amount_to_divide = 0;
 
--- a/math/calculator.cpp
+++ b/math/calculator.cpp
@@ -163,6 +163,14 @@ void requestData() {
             break;
         }
 
+        // Check user's OS
+        // Uses system("cls") if OS is Windows
+        #ifdef __linux__
+        system("clear");
+
+        // Uses system("clear") if OS is Linux
+        #elif _WIN32
         system("cls");
+        #endif
     } while (opt != 5);
 }
--- a/math/calculator.cpp
+++ b/math/calculator.cpp
@@ -8,17 +8,17 @@
 #include <iostream>
 #include <cstdlib>
 
-/** 
+/**
  * Prototypes
  */
-void requestData(float*);
+void requestData(double*);
 void pause();
 
 /**
  * Main function
  */
 int main() {
-    float global_var;
+    double global_var;
     requestData(&global_var);
 
     pause();
@@ -38,8 +38,8 @@ void pause() {
 /**
  * Function to call the calculator.
  */
-void requestData(float *new_val) {
-    int opt = 0; float number;
+void requestData(double *new_val) {
+    int opt = 0; double number;
     int amount_to_sum = 0, amount_to_subtract = 0, amount_to_multiply = 0,
         amount_to_divide = 0;
 
--- a/math/calculator.cpp
+++ b/math/calculator.cpp
@@ -43,7 +43,7 @@ int main() {
  */
 
 void pause() {
-    std::cout << "Press Enter key to continue . . . ";
+    std::cout << "Press ENTER key to continue . . . ";
 
     std::cin.get();
     std::cin.get();
--- a/math/calculator.cpp
+++ b/math/calculator.cpp
@@ -38,7 +38,7 @@ int main() {
      * Release allocated memory used in `new_node`.
      */
 
-    delete[] new_node;
+    delete new_node;
 
     pause();
     return 0;
--- a/math/calculator.cpp
+++ b/math/calculator.cpp
@@ -64,11 +64,10 @@ void requestData() {
                 // Save it in the new node so it's not re-writed
                 if (i == 1) {
                     new_node->datatype = number;
-                }
 
                 // If the current number is not the first one,
                 // sum the current one with the previous one
-                else {
+                } else {
                     new_node->datatype += number;
                 }
             }
@@ -93,11 +92,10 @@ void requestData() {
                 // Save it in the new node so it's not re-writed
                 if (i == 1) {
                     new_node->datatype = number;
-                }
 
                 // If the current number is not the first one,
                 // subtract the current one with the previous one
-                else {
+                } else {
                     new_node->datatype -= number;
                 }
             }
@@ -122,11 +120,10 @@ void requestData() {
                 // Save it in the new node so it's not re-writed
                 if (i == 1) {
                     new_node->datatype = number;
-                }
 
                 // If the current number is not the first one,
                 // multiply the current one with the previous one
-                else {
+                } else {
                     new_node->datatype *= number;
                 }
             }
@@ -151,11 +148,10 @@ void requestData() {
                 // Save it in the new node so it's not re-writed
                 if (i == 1) {
                     new_node->datatype = number;
-                }
 
                 // If the current number is not the first one,
                 // divide the previous number with the current one
-                else {
+                } else {
                     new_node->datatype /= number;
                 }
             }
--- a/math/calculator.cpp
+++ b/math/calculator.cpp
@@ -8,30 +8,18 @@
 #include <iostream>
 #include <cstdlib>
 
-/**
- * Same as the number typed by the user, but it's saved
- * in the new node so it won't be re-writed.
- */
-struct Node {
-    float datatype;
-};
-
 /** 
  * Prototypes
  */
-void requestData(Node*);
+void requestData(float*);
 void pause();
 
 /**
  * Main function
  */
 int main() {
-    Node* new_node = new Node();
-    requestData(new_node);
-
-    // Release allocated memory used in `new_node`.
-
-    delete new_node;
+    float global_var;
+    requestData(&global_var);
 
     pause();
     return 0;
@@ -50,7 +38,7 @@ void pause() {
 /**
  * Function to call the calculator.
  */
-void requestData(Node* new_node) {
+void requestData(float *new_val) {
     int opt = 0; float number;
     int amount_to_sum = 0, amount_to_subtract = 0, amount_to_multiply = 0,
         amount_to_divide = 0;
@@ -81,18 +69,18 @@ void requestData(Node* new_node) {
                 // Save it in the new node so it's not re-writed.
 
                 if (i == 1) {
-                    new_node->datatype = number;
+                    *new_val = number;
 
                     // If the current number is not the first one,
                     // sum the current one with the previous one.
 
                 } else {
-                    new_node->datatype += number;
+                    *new_val += number;
                 }
             }
 
             std::cout << "\nThe sum of the numbers is: "
-                << new_node->datatype << std::endl;
+                << *new_val << std::endl;
 
             pause();
             break;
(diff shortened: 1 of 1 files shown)


math/calculator.cpp
```diff
@@ -0,0 +1,168 @@
+/**
+ * Copyright 2020 @author David Leal
+ * @file
+ *
+ * Basic calculator: sum, subtract, multiply, and divide.
+ */
+
+#include <iostream>
+#include <conio.h>
```
kvedala, line 9: this library is only present in MSVC and not a part of ISO C++ standards - https://en.wikipedia.org/wiki/Conio.h

math/calculator.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include <cstdlib>
+
+struct Node {
```
kvedala, line 12: please document individual functions and objects

math/calculator.cpp
```diff
@@ -0,0 +1,168 @@
...
+int main() {
+    requestData();
+
+    _getch();
+    return 0;
+}
+
+void pause() {
+    std::cout << "Press any key to continue . . . ";
+    _getch();
```
kvedala, line 34: `_getch()` is provided by `conio.h` and not available by GCC or clang and not part of ISO C++ standard

math/calculator.cpp
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            std::cout << "\nThe division of the numbers is: "
+                << new_node->datatype << std::endl;
+
+            pause();
+            break;
+        }
+
+        system("cls");
```
kvedala, line 166: `cls` is a windows command line command and does not work on other OS

Panquesito7: See http://www.cplusplus.com/forum/beginner/73214/.

math/calculator.cpp
```diff
@@ -163,6 +167,15 @@ void requestData() {
...
         }
 
+        // Check user's OS
+        // Uses system("cls") if OS is Windows
+        #ifdef WINDOWS
         system("cls");
+        #endif
+
+        // Uses system("clear") if OS is Linux
+        #ifdef LINUX
```
kvedala, line 177: incorrect ifdef checks-
https://stackoverflow.com/questions/6649936/c-compiling-on-windows-and-linux-ifdef-switch

But, preferably, the "clear" command can be removed altogether. It does not add value to the problem and the code.

kvedala:
Please update comments in the structured format like here - #807

Panquesito7:
> Please update comments

Working on it.

Panquesito7:
> > Please update comments
> 
> Working on it.

Should be done now. 🙂

Panquesito7:
> 1. the cpplint auto checks are still failing - please rectify them

Working on it.

> `conio.h` is not an ISO standard

Working on to use another alternative.

Panquesito7:
@kvedala, all issues have been fixed now.
Thanks for your review and comments! 🙂

math/calculator.cpp
```diff
@@ -0,0 +1,229 @@
...
+};
+
+/** 
+ * Prototypes
+ */
+
+void requestData();
+void pause();
+
+Node* new_node = new Node();
```
kvedala, line 27: avoid using global variables. Create them in the functions and pass as function arguments instead.

Panquesito7: Fixed.

kvedala: > avoid using global variables. Create them in the functions and pass as function arguments instead.
> 
> This will prevent inadvertent bugs. for example, the node here is created as a dynamically allocate memory block __and should be "deleted" otherwise, leads to memory leaks.__

Thank you. The memory allocated using `new` keyword **MUST** be freed/deleted.

Panquesito7: > **MUST** be freed/deleted.

Fixed, allocated memory is now released.
Also added documentation about its use.

math/calculator.cpp
```diff
@@ -34,6 +34,12 @@ int main() {
     Node* new_node = new Node();
     requestData(new_node);
 
+    /**
+     * Release allocated memory used in `new_node`.
+     */
+
+    delete[] new_node;
```
kvedala, line 41: `delete[]` operator is for arrays
`delete` operator is for single arrays
[C++ 101](https://docs.microsoft.com/en-us/cpp/cpp/delete-operator-cpp?view=vs-2019)

Panquesito7: Fixed.

Panquesito7:
Is there anything else that needs to be addressed?

math/calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+ * --------------
+ * Main function
+ * --------------
+ */
+
+int main() {
+    Node* new_node = new Node();
+    requestData(new_node);
+
+    /**
```
kvedala, line 37: normal comments within the code/. Do not use
```
/**
---
 */
```
inside functions

math/calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+    delete new_node;
+
+    pause();
+    return 0;
+}
+
+/**
+ * Similar as system("pause"), but secure.
+ */
+
```
kvedala, line 50: no spaces between comment and function declaration

math/calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+void pause() {
+    std::cout << "Press ENTER key to continue . . . ";
+
+    std::cin.get();
```
kvedala, line 54: only one would be enough

math/calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+            pause();
+            break;
+        }
+
+        /**
+         * Check user's OS
+         * Use system("clear") if OS is GNU/Linux based
+         */
+
+        #ifdef __linux__
```
kvedala, line 221: lines 221-230 can be eliminated. They simply do not add any value to the code and the problem.

math/calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+                } else {
+                    new_node->datatype /= number;
+                }
+            }
+
+            std::cout << "\nThe division of the numbers is: "
+                << new_node->datatype << std::endl;
+
+            pause();
+            break;
```
kvedala, line 213: a `default` case **MUST** always be added

math/calculator.cpp
kvedala: comment line# 17 like below:
```cpp
float datatype; /**< add short description here */
```

math/calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+ */
+
+#include <iostream>
+#include <cstdlib>
+
+/**
+ * Same as the number typed by the user, but it's saved
+ * in the new node so it won't be re-writed.
+ */
+
```
kvedala, line 15: no gap between comment block and the function

math/calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+ * @file
+ *
+ * Basic calculator: sum, subtract, multiply, and divide.
+ */
+
+#include <iostream>
+#include <cstdlib>
+
+/**
+ * Same as the number typed by the user, but it's saved
```
kvedala, line 12: based on the description, even after using the structure, the number will be overwritten. could simply use a variable inside the main function and use pointers to it. Thus avoiding memory allocations altogether and simplifying the program.

Panquesito7: Have you tested the program? It works perfectly for me.
AFAIK, number is not overwritten.

math/calculator.cpp
Panquesito7: Tested with Visual Studio 2019: it doesn't work with one.

math/calculator.cpp
kvedala: unless, I misunderstood,
lines 95, 103, 129, etc each of them updates the value in the node - i.e., the value is overwritten.

kvedala: > Have you tested the program? It works perfectly for me.
> AFAIK, number is not overwritten.

The program would work; but the comment is incorrect and thus rendering the use of node unnecessary.

Panquesito7: How may this be fixed, then?

math/calculator.cpp
Panquesito7: Why shall that be removed? Output would look wrong.

math/calculator.cpp
kvedala: 1. remove struct between line 16-18
2. instead of Node* on line 24 and line 62, replace with `float *new_val`
3. replace all instances of `new_node->datatype` with `*new_val`
4. replace line 34 with `float global_var`
5. replace line 35 with `request_data(&global_var)`
6. remove line 41

math/calculator.cpp
kvedala: there would be no change in the output. This command is only meant to clear the terminal screen. If removed, the results will scroll. 

Please note that this repository is meant for code with algorithms meant for education and not for creating UI elements.

math/calculator.cpp
kvedala: [this is the reason why](https://stackoverflow.com/questions/10656224/using-cin-get-twice)
Essentially, the newline character is not getting flushed and is retained in the `istream` until the first `can.get` is invoked.

math/calculator.cpp
Panquesito7: Should be done now.

math/calculator.cpp
```diff
@@ -8,30 +8,18 @@
...
- */
-struct Node {
-    float datatype;
-};
-
 /** 
  * Prototypes
  */
-void requestData(Node*);
+void requestData(float);
```
kvedala, line 14: use `float*` and not `float`. This should not have compiled.

math/calculator.cpp
```diff
@@ -8,30 +8,18 @@
...
  * Main function
  */
 int main() {
-    Node* new_node = new Node();
-    requestData(new_node);
-
-    // Release allocated memory used in `new_node`.
-
-    delete new_node;
+    float *global_var;
```
kvedala, line 21: here, the variable should be on stack and not a pointer.
Please ensure you run the code before making the commit.

kvedala:
ok. 👍 
@mishraabhinn could you take a look as well. Thanks.

math/calculator.cpp
mishraabhinn: I feel the same.

Panquesito7:
> will go out the range of `int`

First, the datatype is `float`.
Second, IMHO, using the `double` operator may fix this.

Also, can you please re-explain that again?

kvedala:
@mishraabhinn can you run the code for the test case you mentioned?

@Panquesito7  Please remove lines 190-to-197

Panquesito7:
> Please remove lines 190-to-197

Done.

mishraabhinn:
> @mishraabhinn can you run the code for the test case you mentioned?
 
> @Panquesito7 Please remove lines 190-to-197

In general i was asking that if two number whose sum need to be obtained are really big like greater than 10^9 which is above range of int adding both of them can lead to memory overflow. 
And there is no option i can add two number and the sum obtained can be divided by third number. 
There is need of more functions in the program.

kvedala:
I think instead of asking for "number" menu,
the code could simply ask `what operation would you like? (+-*/, ESC to exit): `
The user can then press that. Then it becomes more understandable. 

Also, would using `double` instead resolve your concern, @mishraabhinn

mishraabhinn:
> I think instead of asking for "number" menu,
> the code could simply ask `what operation would you like? (+-*/, ESC to exit): `
> The user can then press that. Then it becomes more understandable.
> 
> Also, would using `double` instead resolve your concern, @mishraabhinn

I agree @kvedala

math/calculator.cpp
Panquesito7: Will use `rewind(stdin)`, as `fflush(stdin)` doesn't work (at least for me).
If requested, I'll use `fflush(stdin)`.

kvedala:
Thinking more about it, I am not sure if this program contributes to any [algorithm](https://en.wikipedia.org/wiki/Algorithm) towards solving a mathematical formulation. This adds, subtracts, multiplies and divides - are they algorithms?

Panquesito7:
@kvedala shall we close this or improve?

kvedala:
> @kvedala shall we close this or improve?

I propose we close and not accept this pr as it simply doesn't introduce any algorithm

PR #805, opened 2020-05-29
Major rework to improve code quality and add automation checks

kvedala:
#### Description of Change

* Add `cmake` based structure 
* Include documentation building
* Add GitHub Actions using `cmake` to check code compilation on linux (ubuntu), Mac OS and Windows (using MS Visual C++ compiler)
* Add GitHub action to auto-generate documentation and push to repo's respective GitHub Pages -> this will create the website https://thealgorithms.github.io/C-Plus-Plus similar to https://kvedala.github.io/C-Plus-Plus
* add documentations to various source codes including mathematical formulations where possible
* fix cpp lint issues with existing files
* fix source codes so they compile on all operating systems
* github actions workflow improved by 
  * adding auto format code and file name
  * stage-by-stage actions for efficient checks
* **New algorithms** - Machine learning and Numerical methods
  * ADALINE learning
  * QR matrix decomposition
  * real eigen value computation using QR decomposition
  * Kohonen Self organizing Maps (SOM) - unsupervised learning
  * roots of n^th degree polynomial using Durand-Kerner algorithm
  * LU decomposition of matrix
  * realtime computation of data statistics - mean, variance and standard deviation
  * ordinary differential equations solver using three Euler methods
* added [gitpod](https://Gitpod.io) integration - with builtin CPPLINT, compile and test and clang-format. Contributors can now simply create a fork, commit and pull their contribution. Since the integration uses VScode, the formatting should be taken care of.

This documentation approach will also enable other contributors to add documentation in their respective languages or use google translate to do so.

~~There is a pull-request (#795) for a smaller subset of the changes made~~ (outdated)
#### Checklist

- [x] Added description of change
- [x] Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTION.md#New-File-Name-guidelines)
- [x] Added tests and example, test must pass
- [x] Relevant documentation/comments is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTION.md#Commit-Guidelines)
- [x] Search previous suggestions before making a new one, as yours may be a duplicate.
- [x] Sort by alphabetical order
- [x] I acknowledge that all my contributions will be made under the project's license.

Notes:

PR #834, opened 2020-06-07
Adding algorithm to check if number is prime or not.

omkarlanghe:
My first contribution to this repository contains algorithm to check if the given number is prime or not.

Small description on how I have optimized the logic:

Reduced all possibilities of a number which cannot be prime.

Eg: No even number, except 2 can be a prime number, hence we will increment our loop with i+2 jumping on all odd numbers only.

If number is <= 1 or if it is even except 2, break the loop and return false telling number is not prime.


kvedala:
1. Use a simpler file name like ‘check_prime.cpp’ 
2. See #807 for how to comment your code. People should be able to read and learn from your code
3. Please done say ‘most optimized algorithm’ if there is no research publication backing your claim

omkarlanghe:
@kvedala Modified the code as per the changes you have suggested.

kvedala:
> @kvedala Modified the code as per the changes you have suggested.

Please see the files changes in #807 - The comment blocks, and structure is per the `doxygen` standard. [link to actual file](https://github.com/TheAlgorithms/C-Plus-Plus/pull/807/files#diff-577d4b488fac5ac13afa9a5d8e229b9b)

GCC suggests - 
```
$ g++ math/check_prime.cpp -o test -Wall
math/check_prime.cpp: In function 'int is_prime(int)':
math/check_prime.cpp:8:23: warning: suggest parentheses around comparison in operand of '&' [-Wparentheses]
    8 |     } else if (no & 1 == 0) {  // otherwise if other even numbers, return false
```

Also, if you want to optimize further, you can have the line# 12 loop condition to go only till `(I*i) < no` - again, this already exists in some of the existing code in the repository.

math/check_prime.cpp
```diff
@@ -1,16 +1,22 @@
...
         return 1;
-    } else if (no & 1 == 0) {  // otherwise if other even numbers, return false
+    } else if ((no & 1) == 0) {
         return 0;
     }
-    if (no >= 3) {  // Start with 3
-        for (int i = 3; i < (no / 2 + 1); i = i + 2) {
-            if (no % i == 0) {
+    if (no >= 3) {
+        for (int i = 3; (i*i) < (no / 2 + 1); i = i + 2) {
```
kvedala, line 18: now this is incorrect - 
should be `(i * i) <= no;`

math/check_prime.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * If number is <= 1 or if it is even except 2, break the loop and return false telling number is not prime.
+ */
+#include <iostream>
+
+ /**
+ * Function to check if the given number is prime or not.
+ * @param no number to be checked.
+ * @return if number is prime, it returns @ true, else it returns @ false.
+ */
+int is_prime(int no) {
```
ayaankhan98, line 14: instead of `no` we can use `num` or just `n`, i think this looks more readable isn't it ?

math/check_prime.cpp
```diff
@@ -0,0 +1,47 @@
...
+    }
+    return (is_prime);
+}
+
+/**
+ * Main function
+ */
+int main() {
+    int n;
+    std::cout << "Enter the number to check if it is prime or not" << std::endl;
```
Panquesito7, line 39: Better readability.
```suggestion
    std::cout << "Enter the number to check if it is prime or not: " << std::endl;
```

omkarlanghe:
Can I close this pull request ?

kvedala:
> Can I close this pull request ?

It will get closed once the pull-request is merged. But for this to happen, the reviewers would have to first approve. 
And for that, could you please make the necessary corrections as requested?
Thank you.

ayaankhan98:
@kvedala is it good enough to merge with master?

math/check_prime.cpp
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright 2020 @author omkarlanghe
```
kvedala, line 2: add `@file` tag in the comments as shown in #807

math/check_prime.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * If number is <= 1 or if it is even except 2, break the loop and return false telling number is not prime.
+ */
+#include <iostream>
+
+ /**
+ * Function to check if the given number is prime or not.
+ * @param num number to be checked.
+ * @return if number is prime, it returns @ true, else it returns @ false.
+ */
+int is_prime(int num) {
```
kvedala, line 14: the function return type should be a bool

omkarlanghe:
@ayaankhan98 Is it okay to modify the code as per @Panquesito7 ?

math/check_prime.cpp
kvedala: Also, could you please use a template for this function-
```cpp
template<typename T>
bool is_prime(T num) {
bool is_prime = true; /* <- this line might have a conflict with the function name. better change it to a con-conflicting variable name */

bool result = true; /* <- maybe this variable would be easy to understand? */
```

math/check_prime.cpp
```diff
@@ -0,0 +1,50 @@
...
+    std::endl;
+    std::cin >> num;
+    int result = is_prime(num);
+    if (result) {
+        std::cout << num << " is a prime number" <<
+        std::endl;
+    } else {
+        std::cout << num << " is not a prime number" <<
+        std::endl;
+    }
```
kvedala, line 49: could you add a test case:

```
assert(is_prime(50) == false);

assert(is_prime(115249) == true);

**Comment by ayaankhan98** [comment]
> @ayaankhan98 Is it okay to modify the code as per @Panquesito7 ?

i will not suggest to modify this current implementaion as per directed by @Panquesito7 because whatever he suggested is a navie approch and is slow, but there is no problem in having both approches in the repository if you wish you can make a PR with this navie approch sepreatly

File: math/check_prime.cpp [hunk 7]
```diff
@@ -0,0 +1,50 @@
...
+    int is_prime = 1;
+    if (num <= 1) {
+        return 0;
+    } else if (num == 2) {
+        return 1;
+    } else if ((num & 1) == 0) {
+        return 0;
+    }
+    if (num >= 3) {
+        for (int i = 3 ; (i*i) < (num) ; i = (i + 2)) {
```
kvedala, line 24: this loop variable should also be the template parameter `T` type.

kvedala:
> > @ayaankhan98 Is it okay to modify the code as per @Panquesito7 ?
> 
> i will not suggest to modify this current implementaion as directed by @Panquesito7 because whatever he suggested is a navie approch and is slow, but there is no problem in having both approches in the repository if you wish you can make a PR with this navie approch sepreatly

Yes, no need to change the algorithm in this implementation. it is good as is.

math/check_prime.cpp
kvedala: > this loop variable should also be the template parameter `T` type.

This change got missed.

kvedala:
when I merge the new master to my fork, my auto tools formatted the code for me 😃 - cdb9766f7a4d3313a7d4aa140268e7b57d16c7a1
and when I reviewed the changes, noticed that the `int main()` function does not return an `int`. The fix will go to #805

Issue #829, opened 2020-06-05
Pythagoras implementation math class.


beqakd:
## Detailed Description
I want to implement some math class function as it was so few in folder. I can start by pythagoras function.

## Context
Its one of the most important feature in math so i think its important.

PaulinaQuintero:
Hello! Can I work on pythagoras function? Or is already taken by someone?

beqakd:
Taken by me. will upload it today.

PaulinaQuintero:
Then I can upload a 3d functionality on pythagoras class

PaulinaQuintero:
I can upload the calculation of the complex's module number to complement that folder

PaulinaQuintero:
I did it
#837 
Could you check it?

PR #835, opened 2020-06-08
Implement pythagoras distance algorithm.

beqakd:
Description of Change
Add implementation of pythagoras distance algorithm. #829 

Checklist
-[ ] Added description of change
 Added file name matches File name guidelines
 Added tests and example, test must pass
 Relevant documentation/comments is changed or added
 PR title follows semantic commit guidelines
 Search previous suggestions before making a new one, as yours may be a duplicate.
 Sort by alphabetical order
 I acknowledge that all my contributions will be made under the project's license.
implement pythagoras distance. style cling format.

### DIFF
--- a/math/pythagoras.cpp
+++ b/math/pythagoras.cpp
@@ -1,20 +1,21 @@
+#include <cassert>
 #include <cmath>
 #include <iostream>
-#include <cassert>
 
 /**
- * The distance between two points in a three dimensional - 3D - coordinate system.
- * For this we have struct point where we save coordinates
- * of our points and later on calculate the distance as mentioned.
+ * The distance between two points in a three dimensional - 3D - coordinate
+ * system. For this we have struct point where we save coordinates of our points
+ * and later on calculate the distance as mentioned.
  */
-struct  Point {
+struct Point {
     int x;
     int y;
     int z;
 };
 
 /**
- * The distance between two points in a three dimensional - 3D - coordinate system
+ * The distance between two points in a three dimensional - 3D - coordinate
+ * system
  */
 int distance(Point a, Point b) {
     int x_2 = std::pow(std::abs(b.x - a.x), 2);
@@ -34,19 +35,18 @@ void tests() {
     point_fir = {0, 0, 7};
     point_sec = {1, 0, 0};
     assert(distance(point_fir, point_sec) == 7);
-    
+
     point_fir = {-7, 0, 7};
     point_sec = {-1, 0, 1};
     assert(distance(point_fir, point_sec) == 8);
-    
+
     point_fir = {100, 200, 300};
     point_sec = {400, 500, 600};
     assert(distance(point_fir, point_sec) == 519);
-    
+
     point_fir = {-1, -2, -3};
     point_sec = {-1, -2, -3};
     assert(distance(point_fir, point_sec) == 0);
-    
 }
 /**
  * Our main function from where we will check our method.
--- a/math/pythagoras.cpp
+++ b/math/pythagoras.cpp
@@ -15,7 +15,7 @@
  * and later on calculate the distance as mentioned.
  */
 class Distance {
-public:
+ public:
     /**
     * Point struct to save 3D coordinates.
     */
@@ -35,8 +35,11 @@ public:
     * @param point_sec 3d coordinates of second point.
     */
     Distance(const int point_fir[], const int point_sec[]) {
-        this -> a = *(const_cast<Point *>(reinterpret_cast<const Point *>(point_fir)));
-        this -> b = *(const_cast<Point *>(reinterpret_cast<const Point *>(point_sec)));
+        Point const *temp1 = reinterpret_cast<const Point *>(point_fir);
+        this->a = *(const_cast<Point *>(temp1));
+
+        Point const *temp2 = reinterpret_cast<const Point *>(point_sec);
+        this->b = *(const_cast<Point *>(temp2));
     }
 
     /**
--- a/math/pythagoras.cpp
+++ b/math/pythagoras.cpp
@@ -5,87 +5,55 @@
  * A basic implementation of 3d distance between two point.
  */
 
-#include <cassert>   // For testing(assert)
-#include <cmath>     // For std::pow(), std::sqrt() and so on
-#include <iostream>  // For io operations
+#include <cassert>  // For testing(assert)
+#include <cmath>    // For std::pow(), std::sqrt() and so on
+#include <iostream> // For io operations
 
 /**
  * The distance between two points in a three dimensional - 3D - coordinate
- * system. For this we have struct point where we save coordinates of our points
- * and later on calculate the distance as mentioned.
+ * system. For this we have 3 point x, y, z.
+ * Also we created method int distance which takes points and returns distance.
  */
-class Distance {
- public:
-    /**
-    * Point struct to save 3D coordinates.
-    */
-    struct Point {
-        int x;
-        int y;
-        int z;
-    };
-
-    // Our two points to calculate distance.
-    Point a;
-    Point b;
-
-    /**
-    * Constructor of this class.
-    * @param point_fir 3D coordinates of first point.
-    * @param point_sec 3d coordinates of second point.
-    */
-    Distance(const int point_fir[], const int point_sec[]) {
-        Point const *temp1 = reinterpret_cast<const Point *>(point_fir);
-        this->a = *(const_cast<Point *>(temp1));
-
-        Point const *temp2 = reinterpret_cast<const Point *>(point_sec);
-        this->b = *(const_cast<Point *>(temp2));
-    }
-
-    /**
-    * This method is just helper one to call private method from public class
-    * and return result of distance.
-    * @return distance between two given 3D point.
-    */
-    double calculateDistance() {
-        double result = distance();
-        return result;
-    }
-
+class point {
  private:
-    /**
-    * The distance between two points in a three dimensional - 3D - coordinate
-    * system. This method is private so it will be called from public class
-    * method.
-    * @return distance between two given 3D point.
-    */
-    int distance() {
-        // initialaize Point a and b
-        Point a = this->a;
-        Point b = this->b;
+    int x;
+    int y;
+    int z;
 
-        // calculate distance and return it.
-        int x_2 = std::pow(std::abs(b.x - a.x), 2);
-        int y_2 = std::pow(std::abs(b.y - a.y), 2);
-        int z_2 = std::pow(std::abs(b.z - a.z), 2);
-        return std::sqrt(x_2 + y_2 + z_2);
-    }
+ public:
+    point(int a, int b, int c)
+        : x(a), y(b), z(c) {}  // Our constructor to fill coordinates.
+    int distance(const point &,
+                const point &);  // Our prototype of distance method.
 };
+
+/**
+ * The distance between two points in a three dimensional - 3D - coordinate
+ * system.
+ * @param point_fir 3D coordinates of first point.
+ * @param point_sec 3d coordinates of second point.
+ * @return returns the distance between this two Point.
+ */
+int point::distance(const point &a, const point &b) {
+    // calculate distance and return it.
+    int x_2 = std::pow(std::abs(b.x - a.x), 2);
+    int y_2 = std::pow(std::abs(b.y - a.y), 2);
+    int z_2 = std::pow(std::abs(b.z - a.z), 2);
+    return std::sqrt(x_2 + y_2 + z_2);
+}
 
 /**
  * In this method we will test distance between 3D coordinates.
  * This one is very basic tests
  */
 void test_1() {
-    int point_fir[3] = {2, 1, 7};
-    int point_sec[3] = {1, 3, 5};
-    Distance dist(point_fir, point_sec);
-    assert(dist.calculateDistance() == 3);
+    point p_fir(2, 1, 7);
+    point p_sec(1, 3, 5);
+    assert(p_fir.distance(p_fir, p_sec) == 3);
 
-    int fir_basic[3] = {0, 0, 7};
-    int sec_basic[3] = {1, 0, 0};
-    Distance dist_basic(fir_basic, sec_basic);
-    assert(dist_basic.calculateDistance() == 7);
+    point fir_basic(0, 0, 7);
+    point sec_basic(1, 0, 0);
+    assert(fir_basic.distance(fir_basic, sec_basic) == 7);
 }
 
 /**
@@ -93,20 +61,17 @@ void test_1() {
  * This one is more advanced tests looking edge cases and other ones.
  */
 void test_2() {
-    int point_fir[3] = {-7, 0, 7};
-    int point_sec[3] = {-1, 0, 1};
-    Distance dist(point_fir, point_sec);
-    assert(dist.calculateDistance() == 8);
+    point p_fir(-7, 0, 7);
+    point p_sec(-1, 0, 1);
+    assert(p_fir.distance(p_fir, p_sec) == 8);
 
-    int fir_high[3] = {100, 200, 300};
-    int sec_high[3] = {400, 500, 600};
-    Distance dist_high(fir_high, sec_high);
-    assert(dist_high.calculateDistance() == 519);
+    point fir_high(100, 200, 300);
+    point sec_high(400, 500, 600);
+    assert(fir_high.distance(fir_high, sec_high) == 519);
 
-    int equal_fir[3] = {-1, -2, -3};
-    int equal_sec[3] = {-1, -2, -3};
-    Distance dist_equal(equal_fir, equal_sec);
-    assert(dist_equal.calculateDistance() == 0);
+    point equal_fir(-1, -2, -3);
+    point equal_sec(-1, -2, -3);
+    assert(equal_fir.distance(equal_fir, equal_sec) == 0);
 }
 
 /**
--- a/math/pythagoras.cpp
+++ b/math/pythagoras.cpp
@@ -3,8 +3,7 @@
 #include <iostream>
 
 // Our class for point coordinates.
-class Point {
-public:
+struct  Point {
     int x;
     int y;
     int z;
--- a/math/pythagoras.cpp
+++ b/math/pythagoras.cpp
@@ -5,9 +5,9 @@
 // Our class for point coordinates.
 class Point {
     public:
-        int x;
-        int y;
-        int z;
+    int x;
+    int y;
+    int z;
 };
 
 // This method just calculate pythagorian distance from a point to b.
--- a/math/pythagoras.cpp
+++ b/math/pythagoras.cpp
@@ -5,9 +5,9 @@
  * A basic implementation of 3d distance between two point.
  */
 
-#include <iostream> // For io operations
-#include <cassert> // For testing(assert)
-#include <cmath> // For std::pow(), std::sqrt() and so on
+#include <cassert>   // For testing(assert)
+#include <cmath>     // For std::pow(), std::sqrt() and so on
+#include <iostream>  // For io operations
 
 /**
  * The distance between two points in a three dimensional - 3D - coordinate
@@ -30,14 +30,13 @@ public:
     Point b;
 
     /**
-    * Constructor of this class. 
+    * Constructor of this class.
     * @param point_fir 3D coordinates of first point.
     * @param point_sec 3d coordinates of second point.
     */
-    Distance(const int point_fir[], const int point_sec[])
-    {
-        this->a = *(Point*)point_fir;
-        this->b = *(Point*)point_sec;
+    Distance(const int point_fir[], const int point_sec[]) {
+        this->a = *(Point *)point_fir;
+        this->b = *(Point *)point_sec;
         ;
     }
 
@@ -46,8 +45,7 @@ public:
     * and return result of distance.
     * @return distance between two given 3D point.
     */
-    double calculateDistance()
-    {
+    double calculateDistance() {
         double result = distance();
         return result;
     }
@@ -59,8 +57,7 @@ private:
     * method.
     * @return distance between two given 3D point.
     */
-    int distance()
-    {
+    int distance() {
         // initialaize Point a and b
         Point a = this->a;
         Point b = this->b;
@@ -77,16 +74,14 @@ private:
  * In this method we will test distance between 3D coordinates.
  * This one is very basic tests
  */
-void test_1()
-{
-
-    int point_fir[3] = { 2, 1, 7 };
-    int point_sec[3] = { 1, 3, 5 };
+void test_1() {
+    int point_fir[3] = {2, 1, 7};
+    int point_sec[3] = {1, 3, 5};
     Distance dist(point_fir, point_sec);
     assert(dist.calculateDistance() == 3);
 
-    int fir_basic[3] = { 0, 0, 7 };
-    int sec_basic[3] = { 1, 0, 0 };
+    int fir_basic[3] = {0, 0, 7};
+    int sec_basic[3] = {1, 0, 0};
     Distance dist_basic(fir_basic, sec_basic);
     assert(dist_basic.calculateDistance() == 7);
 }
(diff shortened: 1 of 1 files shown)


math/pythagoras.cpp
```diff
@@ -0,0 +1,28 @@
+#include <math.h>
+
+#include <iostream>
+
+// Our class for point coordinates.
```
ayaankhan98, line 5: please use doxygen comment style
```
/**
 *
 *
 */
```

math/pythagoras.cpp
```diff
@@ -0,0 +1,28 @@
...
+struct  Point {
+    int x;
+    int y;
+    int z;
+};
+
+// This method just calculate pythagorian distance from a point to b.
```
ayaankhan98, line 12: please use doxygen comment style
```
/**
 *
 *
 */
```

math/pythagoras.cpp
```diff
@@ -0,0 +1,28 @@
+#include <math.h>
```
kvedala, line 1: in C++, use `<cmath>`

math/pythagoras.cpp
```diff
@@ -0,0 +1,28 @@
...
+// Our class for point coordinates.
+struct  Point {
+    int x;
+    int y;
+    int z;
+};
+
+// This method just calculate pythagorian distance from a point to b.
+int distance(Point a, Point b) {
+    int x_2 = pow(abs(b.x - a.x), 2);
```
kvedala, line 14: replace `pow`, `sqrt`, `abs` with their `std::` equivalents
`std::abs` is defined in `<algorithm>` header

math/pythagoras.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+// Our main function.
+int main() {
+    // first point
+    Point point_fir = {2, -1, 7};
+    Point point_sec = {1, -3, 5};
+    // Distance should be 3!
+    std::cout << distance(point_fir, point_sec) << std::endl;
+    return 0;
+}
```
kvedala, line 28: add test cases using `assert` statements.

beqakd:
Will do changes! Thanks a lot.

beqakd:
All changes done!

ayaankhan98:
@beqakd please see #807 this file have a nice documentaion, your implementaion is fine in working perspective do one more thing please update the documentaion too as it is in #807

kvedala:
> @beqakd please see #807 this file have a nice documentaion, your implementaion is fine in working perspective do one more thing please update the documentaion too as it is in #807

Yes. Moreover, this is a pure C implementation and a very good example at that. But, in its current form, it is more appropriate for [this project](https://github.comTheAlgorithms/C). Whereas for this C++ repo, wouldn't it be better to implement `point` as as class and `double point::distance(const point &other)` as its member function? That is the whole point of having C++ - **objects**.

ayaankhan98:
> > @beqakd please see #807 this file have a nice documentaion, your implementaion is fine in working perspective do one more thing please update the documentaion too as it is in #807
> 
> Yes. Moreover, this is a pure C implementation and a very good example at that. But, in its current form, it is more appropriate for [this project](https://github.comTheAlgorithms/C). Whereas for this C++ repo, wouldn't it be better to implement `point` as as class and `double point::distance(const point &other)` as its member function? That is the whole point of having C++ - **objects**.

yes @kvedala you are right, @beqakd please consider these changes also instead of using a `struct` use `class` and write a member function `double point::distance(const point &other)` for this class

kvedala:
Then in that case, @beqakd , 
1. could you make it a C program and create a pull request here - https://github.com/TheAlgorithms/C
2. upgrade the code to C++'s [oop](https://en.wikipedia.org/wiki/Object-oriented_programming) syntax in this PR

Thank you. good code and good exercise 😄

beqakd:
Will do changes. I did struct by looking other your files. So i think we have to create other issue and resolve all of it at once. Most of them were struct even in math folder.

ayaankhan98:
> Will do changes. I did struct by looking other your files. So i think we have to create other issue and resolve all of it at once. Most of them were struct even in math folder.

yes we are in process of resolving those file too soon everything will be upto the point

beqakd:
All changes done! Add documentation, changed to class and fixed cpp format. @ayaankhan98

math/pythagoras.cpp
```diff
@@ -0,0 +1,119 @@
...
+ * The distance between two points in a three dimensional - 3D - coordinate
+ * system. For this we have struct point where we save coordinates of our points
+ * and later on calculate the distance as mentioned.
+ */
+class Distance {
+ public:
+    /**
+    * Point struct to save 3D coordinates.
+    */
+    struct Point {
```
ayaankhan98, line 22: why a `struct` within a `class`?
`class` does not require this, just simply declare the member variables `x`, `y`,  and `z`

math/pythagoras.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <cassert>   // For testing(assert)
+#include <cmath>     // For std::pow(), std::sqrt() and so on
+#include <iostream>  // For io operations
+
+/**
+ * The distance between two points in a three dimensional - 3D - coordinate
+ * system. For this we have struct point where we save coordinates of our points
+ * and later on calculate the distance as mentioned.
+ */
+class Distance {
```
ayaankhan98, line 17: i think we can have a `class point` with a member function which takes two points as it's arguments and return the distance between those points, this will make more sense isn't it?

beqakd: Yes, but to take two point i need struct or something right? So, i will remove struct inside class. Will have variables and will call distance method then.

beqakd: Thats why i had struct in the first place in class so i can call method with 2 Points.

math/pythagoras.cpp
```diff
@@ -0,0 +1,119 @@
...
+    * Point struct to save 3D coordinates.
+    */
+    struct Point {
+        int x;
+        int y;
+        int z;
+    };
+
+    // Our two points to calculate distance.
+    Point a;
```
ayaankhan98, line 29: no need to declare the member variable inside the class just implement the `class point` and create the instances of point class in `test()` to declare two points and call the `point distance(const point &, const point &)` member function on those points this function will return the distance between two points

math/pythagoras.cpp
```diff
@@ -0,0 +1,119 @@
...
+    Point b;
+
+    /**
+    * Constructor of this class.
+    * @param point_fir 3D coordinates of first point.
+    * @param point_sec 3d coordinates of second point.
+    */
+    Distance(const int point_fir[], const int point_sec[]) {
```
ayaankhan98, line 37: why making things too complex when we can achive same functionality with even same speed and accuracy with simple things.

beqakd: Cool. Got your point.  You want to have variable initialized in test method i mean point a and point b right?

beqakd: For this i still need to have Point struct

ayaankhan98:
for your help let me give you a pseudo implementation

```
class point {
private:
    int x;
    int y;
    int z;
public:
    point(int a, int b,  int c): x(a) , y(b) , z(c) {}     // constructor
    point distance(const point &, const point & );
}
point point::distance(const point &p1, const point &p2) {
  // logic goes here
}
void tests() {
// declare points here and write some tests to check the logic
}
int main() {
   test();
   return 0;
}
```
this is just a pseudo implementaion just to give you a idea of what i am suggesting

beqakd:
@ayaankhan98  I think i did changes you asked for. Sorry for inconvenience, this changes and feedback will be very useful for next contributions!

beqakd:
> good job @beqakd thanks for your contribution.

Thanks for helping me out again.

beqakd:
@kvedala Hello, is this pr done or i should change smth.

math/pythagoras.cpp
```diff
@@ -0,0 +1,84 @@
...
+ * Also we created method int distance which takes points and returns distance.
+ */
+class point {
+ private:
+    int x;
+    int y;
+    int z;
+
+ public:
+    point(int a, int b, int c)
```
kvedala, line 24: The constructor function to be documented

math/pythagoras.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <cassert>   // For testing(assert)
+#include <cmath>     // For std::pow(), std::sqrt() and so on
+#include <iostream>  // For io operations
+
+/**
+ * The distance between two points in a three dimensional - 3D - coordinate
+ * system. For this we have 3 point x, y, z.
+ * Also we created method int distance which takes points and returns distance.
+ */
+class point {
```
kvedala, line 17: The class variables can be int, signed, unsigned, float,double, etc. hence, better to template this class

math/pythagoras.cpp
```diff
@@ -0,0 +1,84 @@
...
+ * @param point_fir 3D coordinates of first point.
+ * @param point_sec 3d coordinates of second point.
+ * @return returns the distance between this two Point.
+ */
+int point::distance(const point &a, const point &b) {
+    // calculate distance and return it.
+    int x_2 = std::pow(std::abs(b.x - a.x), 2);
+    int y_2 = std::pow(std::abs(b.y - a.y), 2);
+    int z_2 = std::pow(std::abs(b.z - a.z), 2);
+    return std::sqrt(x_2 + y_2 + z_2);
```
kvedala, line 42: Wouldn’t the sort be non-integer most likely? Return type double would be appropriate

math/pythagoras.cpp
```diff
@@ -0,0 +1,84 @@
...
+};
+
+/**
+ * The distance between two points in a three dimensional - 3D - coordinate
+ * system.
+ * @param point_fir 3D coordinates of first point.
+ * @param point_sec 3d coordinates of second point.
+ * @return returns the distance between this two Point.
+ */
+int point::distance(const point &a, const point &b) {
```
kvedala, line 37: You can overload this function with additional function
```cpp
/** gets distance of current point from the other point (Update this description :) )*/
template<T2>
double point::distance(const point<T2> &other) {
return distance(*this, other);
}
```

math/pythagoras.cpp
```diff
@@ -0,0 +1,84 @@
...
+ * This one is very basic tests
+ */
+void test_1() {
+    point p_fir(2, 1, 7);
+    point p_sec(1, 3, 5);
+    assert(p_fir.distance(p_fir, p_sec) == 3);
+
+    point fir_basic(0, 0, 7);
+    point sec_basic(1, 0, 0);
+    assert(fir_basic.distance(fir_basic, sec_basic) == 7);
```
kvedala, line 56: Using the tee function, this can be written as 
```cpp
assert(fir_basic.distance(sec_basic) == 7);
```

math/pythagoras.cpp
kvedala: You can make this generic by making it 2d as well by setting the third variable default value =0

beqakd: Its good idea.

beqakd:
Nice points. Will change it as fast as possible.

beqakd:
@kvedala  all changes you asked for done. Thanks for points, they were really good.

beqakd:
@ayaankhan98 Can you check this PR.

math/pythagoras.cpp
```diff
@@ -0,0 +1,94 @@
...
+    T y;
+    T z;
+
+ public:
+    /**
+    * We have two constructor. First is for 3D point. it takes T type 3 value
+    * and save it to our calss. Second one takes only 2 T type value and third
+    * one is defoult 0 and save it into the class too.
+    */
+    point(T a, T b, T c) : x(a), y(b), z(c) {}  // Consructor for 3D points
```
kvedala, line 30: in C++11, initializers can be combined with one initializer as such:
```cpp
point(T a, T b, T c=0) : x(a), y(b), z(c) {}  
```
wherein, the third parameter has a default value of zero and becomes an optional input argument.

beqakd: What do you suggest

math/pythagoras.cpp
```diff
@@ -0,0 +1,94 @@
...
+    point(T a, T b) : x(a), y(b), z(0) {}  // Second constructor for 2D points
+    double distance(const point &);  // Our prototype of distance method
```
kvedala, line 32: shouldn't this line be:
```cpp
template<typename T2>
double distance(const point<T2> &other) {
```
notice the `<T2>` for the input argument otherwise, its type could be undefined or implicit typecast?
It is however, compiling using GCC, MSVC and CLANG. wonder what I missed.

math/pythagoras.cpp
kvedala: keeping just one initializer.

kvedala:
> after performing the changes suggested by @kvedala the program will become workable for 2D cartesian coordinate sytem, so i would suggest to modify the current state of docmentation because currently you are considering only 3D coordinate system in documentation prespective

before my suggested changes, program would have worked for 2D as well. there was an explicit 2D initializer. I simply showed to combine to a single way.
I like that way because, people can then easily modify the code to make it take an arbitrary number of arguments and perform the same operation!

math/pythagoras.cpp
```diff
@@ -0,0 +1,94 @@
...
+    T z;
+
+ public:
+    /**
+    * We have two constructor. First is for 3D point. it takes T type 3 value
+    * and save it to our calss. Second one takes only 2 T type value and third
+    * one is defoult 0 and save it into the class too.
+    */
+    point(T a, T b, T c) : x(a), y(b), z(c) {}  // Consructor for 3D points
+    point(T a, T b) : x(a), y(b), z(0) {}  // Second constructor for 2D points
```
ayaankhan98, line 31: i think we do not need this explicitly, because a point in 2D will have Z=0 (always) so we can default z=0 in 3D point constructor by doing this the third argument that is Z becomes optional if the uses passes a 3D point then it will replace Z=0 with the value that is given by user and it will work for the 3D point in this way and if the user does not passes the third argument then it automatically becomes a point in 2D (z=0) and the program will behave accordingly

ayaankhan98:
> > after performing the changes suggested by @kvedala the program will become workable for 2D cartesian coordinate sytem, so i would suggest to modify the current state of docmentation because currently you are considering only 3D coordinate system in documentation prespective
> 
> before my suggested changes, program would have worked for 2D as well. there was an explicit 2D initializer. I simply showed to combine to a single way.
> I like that way because, people can then easily modify the code to make it take an arbitrary number of arguments and perform the same operation!

yeah my  bad luck i had not seen that constructor at that time ;)

kvedala:
Tere is a major issue -> in the tests, the assert statements are converting the output to `int`!! this invalidates the tests altogether.

ayaankhan98:
> Tere is a major issue -> in the tests, the assert statements are converting the output to `int`!! this invalidates the tests altogether.

yes, i don't see a valid reason to cast the distances to `int`. as we are returning `double` after calculation and it seems right to return `double` because distances could be `int` as well as `double`

kvedala:
> > Tere is a major issue -> in the tests, the assert statements are converting the output to `int`!! this invalidates the tests altogether.
> 
> yes, i don't see a valid reason to cast the distances to `int`. as we are returning `double` after calculation and it seems right to return `double` because distances could be `int` as well as `double`

not just that, but the checks would be erroneous, as it does not check the distance values but only the integer close to it!

I think I know why @beqakd used this approach and I can suggest a solution, but I'l wait for the info. 

@ayaankhan98  any update on the #805 ?

ayaankhan98:
> > > Tere is a major issue -> in the tests, the assert statements are converting the output to `int`!! this invalidates the tests altogether.
> > 
> > 
> > yes, i don't see a valid reason to cast the distances to `int`. as we are returning `double` after calculation and it seems right to return `double` because distances could be `int` as well as `double`
> 
> not just that, but the checks would be erroneous, as it does not check the distance values but only the integer close to it!
> 
> I think I know why @beqakd used this approach and I can suggest a solution, but I'l wait for the info.
> 
> @ayaankhan98 any update on the #805 ?

i think he has used this approch to fix the point precision problem?

beqakd:
> > > > Tere is a major issue -> in the tests, the assert statements are converting the output to `int`!! this invalidates the tests altogether.
> 
> > > 
> 
> > > 
> 
> > > yes, i don't see a valid reason to cast the distances to `int`. as we are returning `double` after calculation and it seems right to return `double` because distances could be `int` as well as `double`
> 
> > 
> 
> > not just that, but the checks would be erroneous, as it does not check the distance values but only the integer close to it!
> 
> > 
> 
> > I think I know why @beqakd used this approach and I can suggest a solution, but I'l wait for the info.
> 
> > 
> 
> > @ayaankhan98 any update on the #805 ?
> 
> 
> 
> i think he has used this approch to fix the point precision problem?

Yes, about casting to int. There is problem with asserting to double.

kvedala:
> > > > > Tere is a major issue -> in the tests, the assert statements are converting the output to `int`!! this invalidates the tests altogether.
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > yes, i don't see a valid reason to cast the distances to `int`. as we are returning `double` after calculation and it seems right to return `double` because distances could be `int` as well as `double`
> > 
> > 
> > > 
> > 
> > 
> > > not just that, but the checks would be erroneous, as it does not check the distance values but only the integer close to it!
> > 
> > 
> > > 
> > 
> > 
> > > I think I know why @beqakd used this approach and I can suggest a solution, but I'l wait for the info.
> > 
> > 
> > > 
> > 
> > 
> > > @ayaankhan98 any update on the #805 ?
> > 
> > 
> > i think he has used this approch to fix the point precision problem?
> 
> Yes, about casting to int. There is problem with asserting to double. I can take points if you have better solutions🙌

here is a solution for you: https://kvedala.github.io/C-Plus-Plus/dc/db1/euclidian__distance_8cpp.html

kvedala:
> > > > > > Tere is a major issue -> in the tests, the assert statements are converting the output to `int`!! this invalidates the tests altogether.
> > > 
> > > 
> > > > >
> > > 
> > > 
> > > > >
> > > 
> > > 
> > > > > yes, i don't see a valid reason to cast the distances to `int`. as we are returning `double` after calculation and it seems right to return `double` because distances could be `int` as well as `double`
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > not just that, but the checks would be erroneous, as it does not check the distance values but only the integer close to it!
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > I think I know why @beqakd used this approach and I can suggest a solution, but I'l wait for the info.
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > @ayaankhan98 any update on the #805 ?
> > > 
> > > 
> > > i think he has used this approch to fix the point precision problem?
> > 
> > 
> > Yes, about casting to int. There is problem with asserting to double. I can take points if you have better solutions🙌
> 
> here is a solution for you: https://kvedala.github.io/C-Plus-Plus/dc/db1/euclidian__distance_8cpp.html
> cheers :)

```cpp
/**
 * Use system `epsilon` to compare if the result is close to expected within the
 * accuract limits
 *
 * \param [in] val1 first value to compare
 * \param [in] val2 second value to compare
 * \param [in, optional] accuracy (optional) limit to which to check result in.
 * Default value is system epsilon
 * \returns `true` if values are close within accuracy
 * \returns `false` if value are not equal within accuracy
 */
template <typename T1, typename T2>
bool are_equal(const T1 &val1, const T2 &val2,
               double accuracy = std::numeric_limits<double>::epsilon()) {
    double diff = std::abs(val1 - val2);
    if (diff <= accuracy) {
        return true;
    }

    std::cerr.precision(17);
    std::cerr << " (" << val1 << " - " << val2 << ") = " << diff << " < "
              << accuracy << "\n\t";
    return false;
}

/**
 * In this method we will test distance between coordinates.
 * This one is very basic tests
 */
void self_test() {
    std::cout << "Test 1...";
    point<int> p_fir({2, 1, 7});
    point<int> p_sec({1, 3, 5});
    assert(p_fir.distance(p_sec) == 3);
    std::cout << "passed!\n";

    std::cout << "Test 2...";
    point<int> fir_basic({0, 0, 7});
    point<int> sec_basic({1, 0, 0});
    assert(are_equal(fir_basic.distance(sec_basic), 7.0710678118654755));
    std::cout << "passed!\n";

    std::cout << "Test 3...";
    point<int> fir_err({0, 0, 7});
    point<int> sec_err({1, 0});
    assert(fir_err.distance(sec_err) < 0);
    std::cout << "passed!\n";

    std::cout << "Test 4...";
    point<double> p_fir4({-7.5, 7.2});
    point<double> p_sec4({-1.0, 1.5});
    assert(are_equal(p_fir4.distance(p_sec4), 8.6452298986203946));
    std::cout << "passed!\n";

    std::cout << "Test 5...";
    point<int64_t> fir_high({100, 200, 300});
    point<int64_t> sec_high({400, 500, 600});
    assert(are_equal(fir_high.distance(sec_high), 519.6152422706632));
    std::cout << "passed!\n";

    std::cout << "Test 6...";
    point<int16_t> equal_fir({-1, -2, -3});
    point<int16_t> equal_sec({-1, -2, -3});
    assert(are_equal(equal_fir.distance(equal_sec), 0));
    std::cout << "passed!\n";

    std::cout << "Test 7...";
    point<float> two_fir({-1.9, -2.5});
    point<float> two_sec({-19.2, -10.5});
    assert(are_equal(two_fir.distance(two_sec), 19.060168914847004));
    std::cout << "passed!\n";
}
```

math/pythagoras.cpp
beqakd: Its not if you declare method outside of the class like this
template <typename T> double point<T>::distance(const point &other)
So its not problem.

beqakd:
Thanks for all of your points. @kvedala I changed constructor and also thanks for tests. Will be helpful in future contributions!

math/pythagoras.cpp
```diff
@@ -0,0 +1,131 @@
...
+    point<int> fir_basic({0, 0, 7});
+    point<int> sec_basic({1, 0, 0});
+
+    assert(are_equal(fir_basic.distance(sec_basic), 7.0710678118654755));
+    std::cout << "passed!\n";
+
+    std::cout << "Test 3...";
+    point<int> fir_err({0, 0, 7});
+    point<int> sec_err({1, 0});
+    assert(are_equal(fir_basic.distance(sec_basic), 7.0710678118654755));
```
kvedala, line 97: this line is an incorrect check

beqakd: Okey, i will throw exception somehow

beqakd: It goes for default 0 value. Thats why it doesnot fails.

kvedala: > It goes for default 0 value. Thats why it doesnot fails.

Thinking about it, the check is fine. Line# 96 will still have the `sec_err` as 2D point but with `z=0`. The result should be a non-zero distance 
formula^2%2B(0-0)^2%2B(7-0)^2}=\sqrt{8}=2\sqrt{2}\approx2.8284271247)

beqakd: Yes,agreed.

beqakd: Will check it

beqakd: Everything is clear there you forgot 7^2 + 1 its 50 not 8 @kvedala

beqakd: So its clear.

kvedala:
@beqakd Once you are done with the fix, we can merge your contribution.

beqakd:
> @beqakd Once you are done with the fix, we can merge your contribution.

I am still waiting your reply. This test was correct you forgot to square 7. Check up comment

kvedala:
> > @beqakd Once you are done with the fix, we can merge your contribution.
> 
> I am still waiting your reply. This test was correct you forgot to square 7. Check up comment. @kvedala

Oh yes! Thanks for clarifying.
@ayaankhan98 This can be merged

ayaankhan98:
```
/home/runner/work/C-Plus-Plus/C-Plus-Plus/math/pythagoras.cpp:65:39: error: ‘numeric_limits’ is not a member of ‘std’
                double accuracy = std::numeric_limits<double>::epsilon()) {
                                       ^~~~~~~~~~~~~~
/home/runner/work/C-Plus-Plus/C-Plus-Plus/math/pythagoras.cpp:65:54: error: expected primary-expression before ‘double’
                double accuracy = std::numeric_limits<double>::epsilon()) {
                                                      ^~~~~~
/home/runner/work/C-Plus-Plus/C-Plus-Plus/math/pythagoras.cpp:65:54: error: expected ‘,’ or ‘...’ before ‘double’
```
compile check Awesome CI build failure on ubuntu

beqakd:
> ```
> 
> /home/runner/work/C-Plus-Plus/C-Plus-Plus/math/pythagoras.cpp:65:39: error: ‘numeric_limits’ is not a member of ‘std’
> 
>                 double accuracy = std::numeric_limits<double>::epsilon()) {
> 
>                                        ^~~~~~~~~~~~~~
> 
> /home/runner/work/C-Plus-Plus/C-Plus-Plus/math/pythagoras.cpp:65:54: error: expected primary-expression before ‘double’
> 
>                 double accuracy = std::numeric_limits<double>::epsilon()) {
> 
>                                                       ^~~~~~
> 
> /home/runner/work/C-Plus-Plus/C-Plus-Plus/math/pythagoras.cpp:65:54: error: expected ‘,’ or ‘...’ before ‘double’
> 
> ```
> 
> compile check Awesome CI build failure on ubuntu
> 
> 
> 
> can be fixed by adding `#include <limits>`, `std::numeric_limits` is defined in `limits`

Will check right ahead!

PR #844, opened 2020-06-11
Improving LinkedList implementation

ananyasharma1442:
As mentioned in the guidelines, I have changed the Struct name to Class, though it will decently increase the compile time. Class is preffered as we are using the c++ language.

### DIFF
--- a/data_structure/linked_list.cpp
+++ b/data_structure/linked_list.cpp
@@ -1,8 +1,9 @@
 #include <iostream>
 
-struct node {
-        int val;
-        node *next;
+class node {
+        public:
+            int val;
+            node *next;
 };
 
 node *start;
@@ -16,11 +17,11 @@ void insert(int x) {
                 while (t->next != NULL) {
                         t = t->next;
                 }
-                t->next = n;             
+                t->next = n;
         } else {
                 start = n;
         }
-        
+
 }
 
 void remove(int x) {
--- a/data_structure/linked_list.cpp
+++ b/data_structure/linked_list.cpp
@@ -1,107 +1,115 @@
 #include <iostream>
-
+template  <class X>
 class node {
-        public:
-            int val;
-            node *next;
+public:
+    X val;
+    node<X> *next;
 };
 
+
+template  <class X>
 class linked_list
 {
 private:
-    node *start;
+    node<X> *start;
 public:
     linked_list()
     {
         start=NULL;
     }
-
-    void insert(int x) {
-            node *t = start;
-            node *n = new node;
-            n->val = x;
-            n->next = NULL;
-            if (start != NULL) {
-                    while (t->next != NULL) {
-                            t = t->next;
-                    }
-                    t->next = n;
-            } else {
-                    start = n;
-            }
-    }
-
-    void remove(int x) {
-            if (start == NULL) {
-                    std::cout << "\nLinked List is empty\n";
-                    return;
-            } else if (start->val == x) {
-                    node *temp = start;
-                    start = start->next;
-                    delete temp;
-                    return;
-            }
-            node *temp = start, *parent = start;
-            while (temp != NULL && temp->val != x) {
-                    parent = temp;
-                    temp = temp->next;
-            }
-            if (temp == NULL) {
-                    std::cout << std::endl << x << " not found in list\n";
-                    return;
-            }
-            parent->next = temp->next;
-            delete temp;
-    }
-
-        void search(int x) {
-            node *t = start;
-            int found = 0;
-            while (t != NULL) {
-                    if (t->val == x) {
-                            std::cout << "\nFound";
-                            found = 1;
-                            break;
-                    }
-                    t = t->next;
-            }
-            if (found == 0) {
-                    std::cout << "\nNot Found";
-            }
-    }
-
-    void show() {
-            node *t = start;
-            while (t != NULL) {
-                    std::cout << t->val << "\t";
-                    t = t->next;
-            }
-    }
-
-    void reverse() {
-            node *first = start;
-            if (first != NULL) {
-                    node *second = first->next;
-                    while (second != NULL) {
-                            node *tem = second->next;
-                            second->next = first;
-                            first = second;
-                            second = tem;
-                    }
-                    start->next = NULL;
-                    start = first;
-            } else {
-                    std::cout << "\nEmpty list";
-            }
-    }
+    void insert (X);
+    void remove (X);
+    void search (X);
+    void show();
+    void reverse();
 };
 
 
+template  <class X>
+void linked_list<X>::insert(X x) {
+        node <X>*t = start;
+        node <X>*n = new node<X>;
+        n->val = x;
+        n->next = NULL;
+        if (start != NULL) {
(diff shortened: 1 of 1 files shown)


ayaankhan98:
@ananyasharma1442 one more suggestion is here try to update this implementaion to a generic level using templates.

ayaankhan98:
@ananyasharma1442 please see #807 it has a nice documentation, do the documentation of your code in same manner

data_structure/linked_list.cpp
```diff
@@ -1,57 +1,71 @@
...
-struct node {
-        int val;
-        node *next;
+template  <class X>
+class linked_list
+{
+private:
+    node<X> *start;
+public:
+    linked_list()
```
ayaankhan98, line 16: a destructor is also required i think.

data_structure/linked_list.cpp
```diff
@@ -106,26 +123,26 @@ int main() {
...
                         std::cout << "\n";
                         break;
                  case 5:
                         std::cout << "The reversed list: \n";
-                        reverse();
-                        show();
+                        object.reverse();
+                        object.show();
                         std::cout << "\n";
                         break;
```
mishraabhinn, line 147: A `default:` case is always good to have.

data_structure/linked_list.cpp
```diff
@@ -66,33 +80,36 @@ void search(int x) {
...
+    }
 }
 
+
 int main() {
         int choice, x;
+        linked_list<int> object;
         do {
                 std::cout << "\n1. Insert";
                 std::cout << "\n2. Delete";
```
mishraabhinn, line 115: Add `endl;` after each statement.

data_structure/linked_list.cpp
```diff
@@ -66,33 +80,36 @@ void search(int x) {
...
+            while (second != NULL) {
+                    node<X> *tem = second->next;
+                    second->next = first;
+                    first = second;
+                    second = tem;
+            }
+            start->next = NULL;
+            start = first;
+    } else {
+            std::cout << "\nEmpty list";
```
mishraabhinn, line 105: Add `endl;` at the statement end.

data_structure/linked_list.cpp
ananyasharma1442: Adding "endl" would have no meaning if we have "\n" at the beginning of the statements. Running it on my console just prints them perfectly.

mishraabhinn: Agreed. Bu I feel using `endl;`is much better than `"\n";`.
<a href="https://www.geeksforgeeks.org/endl-vs-n-in-cpp/#:~:text=endl%20vs%20%5Cn%20in%20C%2B%2B,-Although%20they%20both&text=cout%20%3C%3C%20endl%20%3A%20Inserts%20a,Only%20inserts%20a%20new%20line.&text=cout%20%3C%3C%20'%5Cn',flushing%20of%20stream%20is%20required.">**Refer**</a>
So whenever dynamic allocation is done, It's always better to use `endl;` instead of `"\n";`

data_structure/linked_list.cpp
kvedala: agreed - for memory safety.

data_structure/linked_list.cpp
ayaankhan98: i personally prefer to use '\n' because it is fast as it does not flushes the stream, but using '\n' statements become a bit unclear in reading

mishraabhinn: I don't have problem with "\n"
But  in this program use fflush(stdout) and fflush(stdin) to clear buffer sometimes wrong output may come due to unclear buffer. 
<a href="https://www.geeksforgeeks.org/clearing-the-input-buffer-in-cc/">Refer</a>

data_structure/linked_list.cpp
```diff
@@ -1,57 +1,71 @@
 #include <iostream>
+template  <class X>
+class node {
+public:
+    X val;
+    node<X> *next;
```
kvedala, line 6: adding a constructor and destructor would help - otherwise, this is a plain  structure.
` node(X value) (val: value) {;}`
should do.

A recursive destructor - to ensure clean up of the `*next` on exit

data_structure/linked_list.cpp
ayaankhan98: yes you are right @mishraabhinn that in case of char arrays or strings '\n' creates a problem if we do not remove '\n' form the buffer or in other words it becomes necessary to flush the buffer but this all happens when we use C function `scanf()` or in C++ if we use something like `cin.getline()` but in this implementaion we are not using any of these functions we are taking input with `std::cin` only so i think it will not cause such buffer issues. well i don't have any problem in using `std::endl`.

@ananyasharma1442 please replace '\n' with `std::endl`

ananyasharma1442:
**Description of Change**
From the previous implementation, I have changed the struct to class and all the class functions preside in the class itself. Also used templates (generics) to make the program easier to read and functional.

**Checklist**

- [x] Added description of the change
- [x] Added file name matches File name guidelines
- [x] PR title follows semantic commit 

ayaankhan98:
@ananyasharma1442 cpplint is failing 
```
Done processing data_structure/linked_list.cpp
Total errors found: 11
Python 3.8.3
1 files were modified.
1 C++ files were modified.
cpplint:
data_structure/linked_list.cpp:12:  Weird number of spaces at line-start.  Are you using a 2-space indent?  [whitespace/indent] [3]
data_structure/linked_list.cpp:12:  { should almost always be at the end of the previous line  [whitespace/braces] [4]
data_structure/linked_list.cpp:16:  Missing space before {  [whitespace/braces] [5]
data_structure/linked_list.cpp:26:  { should almost always be at the end of the previous line  [whitespace/braces] [4]
data_structure/linked_list.cpp:27:  Redundant blank line at the start of a code block should be deleted.  [whitespace/blank_line] [2]
data_structure/linked_list.cpp:27:  Redundant blank line at the end of a code block should be deleted.  [whitespace/blank_line] [3]
data_structure/linked_list.cpp:29:  Redundant blank line at the end of a code block should be deleted.  [whitespace/blank_line] [3]
data_structure/linked_list.cpp:154:  Missing spaces around !=  [whitespace/operators] [3]
data_structure/linked_list.cpp:154:  Missing space before ( in if(  [whitespace/parens] [5]
data_structure/linked_list.cpp:154:  Missing space before {  [whitespace/braces] [5]
data_structure/linked_list.cpp:155:  Missing spaces around <<  [whitespace/operators] [3]
Traceback (most recent call last):
  File "/home/runner/work/_temp/41a8a809-50d0-4ccc-8004-ac16072724b2.py", line 17, in <module>
    subprocess.run(["cpplint", "--filter=-legal/copyright"] + cpp_files, check=True, text=True)
  File "/opt/hostedtoolcache/Python/3.8.3/x64/lib/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['cpplint', '--filter=-legal/copyright', 'data_structure/linked_list.cpp']' returned non-zero exit status 1.
```

PR #846, opened 2020-06-14
A program to check if given number is odd or even without using any arithmetic operator.

omkarlanghe:
- A program to check if given number is odd or even without using any arithmetic operator.
- This program illustrates the basic use of bitwise AND to check if number is odd or even. 
- I will try to add more programs related to bitwise operators to help other learner's to understand how to play with bits.
- I also suggest other contributor's to add programs to help understand how bitwise operators works.
- For this, I have created a separate folder name **bitwise_operators** to keep things simple to search.


kvedala:
The even odd function is Too trivial. The Prime check program is identical to #834

omkarlanghe:
> The even odd function is Too trivial. The Prime check program is identical to #834

Should I make it into main function itself ?

kvedala:
> > The even odd function is Too trivial. The Prime check program is identical to #834
> 
> Should I make it into main function itself ?

No. I don’t think the even-odd function adds “algorithmic” value. It is a binary logic trick and not an “algorithm”.

bitwise_operators/check_even_odd.cpp
```diff
@@ -0,0 +1,71 @@
...
+    if (result) {
+        std::cout << "The given number " << num << " is even." <<
+        std::endl;
+    } else {
+        std::cout << "The given number " << num << " is odd." <<
+        std::endl;
+    }
+
+    assert(is_odd_even(4) == true);
+    assert(is_odd_even(7) == false);
```
Panquesito7, line 70: Please add `return 0;` at the end of the program
It is a good programming practice.

math/check_prime.cpp
```diff
@@ -0,0 +1,58 @@
...
+    bool result = is_prime(num);
+    if (result) {
+        std::cout << num << " is a prime number" <<
+        std::endl;
+    } else {
+        std::cout << num << " is not a prime number" <<
+        std::endl;
+    }
+    assert(is_prime(50) == false);
+    assert(is_prime(115249) == true);
```
Panquesito7, line 57: Same here.

bitwise_operators/check_even_odd.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <iostream>
+#include <cassert>
+
+/**
+ * Function to check if the given number is odd or even.
+ * @param num number to be checked.
+ * @return if number is even, it returns @ true, else it returns @ false.
+*/
+template<typename T>
+bool is_odd_even(T num) {
```
tjgurwara99, line 43: I think this operation is too trivial to be in a fully fledged cpp file. The trick of using bitwise operation for c++ users is very common.
For example:
n & 0x1 checks for even and odd (essentially what you did) and 
n >> 0x1 right shifts the binary (essentially divides by 2) 
and there are quite a few more bitwise operations that we use and are very common for c++ users.

bitwise_operators/check_even_odd.cpp
```diff
@@ -0,0 +1,71 @@
...
+    if ((num & 1) == 0) {
```
tjgurwara99, line 44: I think this operation is too trivial to be in a fully fledged cpp file. The trick of using bitwise operation for c++ users is very common.
For example:
n & 0x1 checks for even and odd (essentially what you did) and
n >> 0x1 right shifts the binary (essentially divides by 2)
and there are quite a few more bitwise operations that we use and are very common for c++ users.

PR #852, opened 2020-06-15
Babylonian method for square root

Anmolreshi:
#include <iostream> 
using namespace std; 
class root { 
    /*Returns the square root of n. Note that the function */
public: 
    float squareRoot(float n) 
    { 
        /*We are using n itself as initial approximation 
          This can definitely be improved */
        float x = n; 
        float y = 1; 
        float e = 0.000001; /* e decides the accuracy level*/
        while (x - y > e) { 
            x = (x + y) / 2; 
            y = n / x; 
        } 
        return x; 
    } 
}; 
  
/* Driver program to test above function*/
int main() 
{ 
    root g; 
    int n ;
cin>>n; 
    cout << "Square root of " << n << " is " << g.squareRoot(n); 
    getchar(); 
}


math/sqrt_double.cpp
```diff
@@ -6,6 +6,9 @@ number in O(logn) time,
 with precision fixed */
 
 double Sqrt(double x) {
+    if ( x > 0 && x < 1 ) {
```
kvedala, line 9: This is the only part that is new. 

The pull-request is invalid. Please see some tutorials on how to commit and create a pull-request - [example link](https://opensource.com/article/19/7/create-pull-request-github) 

If you are contributing, contribute a new file and not change existing one unless you are making any corrections to them. 

Thank you

ayaankhan98:
@Anmolreshi would you please create a seprate pull request from your fork with this single file `babylonian_method_sqrt.cpp`.

PR #873, opened 2020-06-20
[Bug Fix] of sorting/heapsort.cpp

ayaankhan98:
#### Description of Change
Fresh implementation of heapsort

#### Checklist
- Added description of change
- Added tests and example, test must pass
- Added documentation so that the program is self-explanatory and educational - [Doxygen guidelines](https://www.doxygen.nl/manual/docblocks.html)
- Relevant documentation/comments is changed or added
- I acknowledge that all my contributions will be made under the project's license.


kvedala:
> #### Description of Change
> Fresh implementation of heapsort
> 
> #### Checklist
> * Added description of change
> * Added tests and example, test must pass
> * Added documentation so that the program is self-explanatory and educational - [Doxygen guidelines](https://www.doxygen.nl/manual/docblocks.html)
> * Relevant documentation/comments is changed or added
> * I acknowledge that all my contributions will be made under the project's license.

1. any particular reason to use lambda functions? They should be small - typically one line returns that will be inlined at compile-time. Otherwise, the performance can actually be hampered. Moreover, it affects the readability of the code. Having a loop in lambda functions is, typically, a bad idea.
2. in the documentation for the heaping function, the parameters are not documented.

ayaankhan98:
there is no particular reason to use lambda functions, but in regard with performance i earlier thought this that it may affect the performance so i discussed the issue with some of the members of Cpp lang community they just said to me that their is no difference in performance of normal functions and lambda functions

well that's not a big issue here i will fix this soon :)

kvedala:
> there is no particular reason to use lambda functions, but in regard with performance i earlier thought this that it may affect the performance so i discussed the issue with some of the members of Cpp lang community they just said to me that their is no difference in performance of normal functions and lambda functions
> 
> well that's not a big issue here i will fix this soon :)

A good example where could be useful is in the numerical-methods - For example, the root-finding and interpolation algorithms like Newton-Raphson, Durand-Kerner, etc that evaluate a function at a value. You can create new implementations of these programs showing the application of lambda functions and how it improves the code.

ayaankhan98:
> > there is no particular reason to use lambda functions, but in regard with performance i earlier thought this that it may affect the performance so i discussed the issue with some of the members of Cpp lang community they just said to me that their is no difference in performance of normal functions and lambda functions
> > well that's not a big issue here i will fix this soon :)
> 
> A good example where could be useful is in the numerical-methods - For example, the root-finding and interpolation algorithms like Newton-Raphson, Durand-Kerner, etc that evaluate a function at a value. You can create new implementations of these programs showing the application of lambda functions and how it improves the code.

yes you are right they are pretty useless in this case, the only reason i used them here is just to give a taste of lambdas to the reader for educational purpose, as i feel most of the people are unaware of these concepts of cpp.

well it's a good suggestion i will try to implement numerical methods using lambdas

kvedala:
> > > there is no particular reason to use lambda functions, but in regard with performance i earlier thought this that it may affect the performance so i discussed the issue with some of the members of Cpp lang community they just said to me that their is no difference in performance of normal functions and lambda functions
> > > well that's not a big issue here i will fix this soon :)
> > 
> > 
> > A good example where could be useful is in the numerical-methods - For example, the root-finding and interpolation algorithms like Newton-Raphson, Durand-Kerner, etc that evaluate a function at a value. You can create new implementations of these programs showing the application of lambda functions and how it improves the code.
> 
> yes you are right they are pretty useless in this case, the only reason i used them here is just to give a taste of lambdas to the reader for educational purpose, as i feel most of the people are unaware of these concepts of cpp.
> 
> well it's a good suggestion i will try to implement numerical methods using lambdas

👍 also, refer the issue in the description of this pull request

ayaankhan98:
@kvedala review and let me know if any other changes are required

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,96 @@
...
-}
+/**
+ * heapSort lambda function utilizes heapify procedure to sort
+ * the array
+ *
+ * @param arr array to be sorted
+ * @param n size of array
+ *
+ */
+void heapSort(int *arr, int n) {
```
kvedala, line 78:
```suggestion
template<typename T>
void heapSort(T *arr, int n) {
```

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,96 @@
...
+ * the bottom up by successively sifting downward to establish the
+ * heap property.
+ *
+ * @param arr array to be sorted
+ * @param n size of array
+ * @param i node position in Binary Tress or element position in
+ *          Array to be compared with it's childern
+ *
+ */
+void heapify(int *arr, int n, int i) {
```
kvedala, line 53:
```suggestion
template <typename T>
void heapify(T *arr, int n, int i) {
```

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,96 @@
...
+/**
+ *
+ * Utility Lambda function to print the array after
+ * sorting.
+ *
+ * @param arr array to be printed
+ * @param sz size of array
+ *
+ */
+void printArray(int *arr, int sz) {
```
kvedala, line 36:
```suggestion
template <typename T>
void printArray(T *arr, int sz) {
```

kvedala: > Suggested change 
> void printArray(int *arr, int sz) {
> template <typename T>
> void printArray(T *arr, int sz) {

Should this function be in camel case `print_array`?

kvedala: > > Suggested change
> > void printArray(int *arr, int sz) {
> > template 
> > void printArray(T *arr, int sz) {
> 
> Should this function be in snake case `print_array`?

Never mind. In c++ google style specs, this is allowed

sorting/heap_sort.cpp
```diff
@@ -96,20 +96,20 @@ void test() {
...
     double arr2[] = {4.5, -3.6, 7.6, 0, 12.9};
     sz = sizeof(arr2) / sizeof(arr2[0]);
     printArray(arr2, sz);
     heapSort(arr2, sz);
     printArray(arr2, sz);
-    assert(std::is_sorted(arr2, arr2+sz));
+    assert(std::is_sorted(arr2, arr2 + sz));
     std::cout << "Test 2 passed\n";
 }
 /** Main function */
```
kvedala, line 112:
```suggestion

/** Main function */
```

sorting/heap_sort.cpp
ayaankhan98: okay if it is allowed then i am leaving this in camel casing

sorting/heap_sort.cpp
```diff
@@ -96,20 +96,20 @@ void test() {
     printArray(arr, sz);  // displaying the array before sorting
     heapSort(arr, sz);    // calling heapsort to sort the array
     printArray(arr, sz);  // display array after sorting
-    assert(std::is_sorted(arr, arr+sz));
+    assert(std::is_sorted(arr, arr + sz));
     std::cout << "Test 1 Passed\n========================\n";
     std::cout << "Test 2\n";
```
kvedala, line 102:
```suggestion

    std::cout << "Test 2\n";
```

kvedala:
Good job 👍

ayaankhan98:
> Good job

are we good to merge this with master?

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,118 @@
+/**
+ * \addtogroup sorting Sorting Algorithm
+ * @{
+ * \file
+ * \brief [Heap Sort Algorithm
+ * (HEAP SORT)](https://en.wikipedia.org/wiki/Heapsort) implementation
```
kvedala, line 6:
```suggestion
 * (heap sort)](https://en.wikipedia.org/wiki/Heapsort) implementation
```
no need of all caps

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,118 @@
...
+ *
+ * \author [Ayaan Khan](http://github.com/ayaankhan98)
+ *
+ * \details
+ *  heapsort is a comparison-based sorting algorithm.
```
kvedala, line 11:
```suggestion
 *  Heap-sort is a comparison-based sorting algorithm.
```

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,118 @@
...
+ *  like selection sort, heapsort divides its input into a sorted
+ *  and an unsorted region, and it iteratively shrinks the unsorted
+ *  region by extracting the largest element from it and inserting
+ *  it into the sorted region. Unlike selection sort,
+ *  heapsort does not waste time with a linear-time scan of the
+ *  unsorted region; rather, heap sort maintains the unsorted region
+ *  in a heap data structure to more quickly find the largest element
+ *  in each step.
+ *
+ *  Time Complexity - O(nlog(n))
```
kvedala, line 22:
```suggestion
 *  Time Complexity - \f$O(n \log(n))\f$
```

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,118 @@
...
 }
 
-void heapsort(int *a, int n) {
-    for (int i = n - 1; i >= 0; --i) {
-        std::swap(a[0], a[i]);
-        heapify(a, 0, i);
-    }
-}
+/**
+ * heapSort lambda function utilizes heapify procedure to sort
```
kvedala, line 75:
```suggestion
 * Utilizes heapify procedure to sort
```

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,118 @@
+/**
+ * \addtogroup sorting Sorting Algorithm
```
kvedala, line 2:
```suggestion
```
since the heal sort algorithm is essentially only two files, I suggest the group should contain only these files and not the test and main functions

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,118 @@
...
+ * @{
```
kvedala, line 3:
```suggestion
```

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,118 @@
...
+ * @param sz size of array
+ *
+ */
+template <typename T>
+void printArray(T *arr, int sz) {
+    for (int i = 0; i < sz; i++) std::cout << arr[i] << "  ";
+    std::cout << "\n";
+}
+
+/**
```
kvedala, line 44:
```suggestion
/**
 * \addtogroup sorting Sorting Algorithm
 * @{
```

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,118 @@
...
+    for (int i = n - 1; i >= 0; i--) heapify(arr, n, i);
 
-void build_maxheap(int *a, int n) {
-    for (int i = n / 2 - 1; i >= 0; --i) {
-        heapify(a, i, n);
+    for (int i = n - 1; i >= 0; i--) {
+        std::swap(arr[0], arr[i]);
+        heapify(arr, i, 0);
     }
 }
```
kvedala, line 91:
```suggestion
/**
 * @}
 */
 
```

sorting/heap_sort.cpp
```diff
@@ -1,52 +1,118 @@
...
+    std::cout << "Test 2 passed\n";
+}
 
-    std::getchar();
+/** Main function */
+int main() {
+    test();
+    return 0;
 }
+/** @} */
```
kvedala, line 118:
```suggestion
```

PR #875, opened 2020-06-21
Given an unsorted array A of size N of non-negative integers, find a continuous sub-array which adds to a given number S, If not found print -1.

omkarlanghe:
Given an unsorted array A of size N of non-negative integers, find a continuous sub-array which adds to a given number S, If not found print -1.


others/sub_array_with_given_sum.cpp
```diff
@@ -0,0 +1,98 @@
...
+        std::cout << "Enter the elements in an array : " << std::endl;
+        for (int i = 0; i < size; i++) {
+            std::cin >> arr[i];
+        }
+
+        continues_sub_array(arr, size, orignal_sum);
+
+        std::cout << std::endl;
+        // free memory of an array once used.
+        delete arr;
```
Panquesito7, line 95: Incorrect, please use:
 ```c++
delete[] arr
```

File: others/sub_array_with_given_sum.cpp [hunk 2]
```diff
@@ -0,0 +1,98 @@
...
+    }
+    return (0);
```
Panquesito7, line 97: No need to use parenthesis here.
```suggestion
    return 0;
```

omkarlanghe: done

bitwise_operators/check_even_odd.cpp
```diff
@@ -0,0 +1,73 @@
+/**
```
kvedala, line 1: as discussed in reviews above, this file should be removed from this pull-request

others/sub_array_with_given_sum.cpp
```diff
@@ -0,0 +1,97 @@
...
+ * Function to print the index range of subarray equal to the given sum. If no
+ * such subarray found, print -1.
+ * @param arr array to check
+ * @param size size of an array
+ * @param original_sum sum to match with the sum of all elements of a sub array.
+ * @return void and prints index ranges if sum of sub array equal to the given
+ * sum, else prints -1.
+ */
+template <typename T>
+void contiguous_sub_array(T *arr, T size, T orignal_sum) {
```
kvedala, line 39: 1. size of array is represented in `size_t` and not `T`
2. Template would not be useful in this particular program.

others/sub_array_with_given_sum.cpp
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Main function
+ */
+int main() {
+    int test_cases, size, orignal_sum;
+
+    std::cout << "Enter test cases : " << std::endl;
+    std::cin >> test_cases;
+
+    while (test_cases--) {
```
kvedala, line 75: implement self-test cases as in #873

others/sub_array_with_given_sum.cpp
```diff
@@ -0,0 +1,89 @@
+/**
+ * Copyright 2020 @author omkarlanghe
+ *
+ * @file
+ * Given an unsorted array A of size N of non-negative integers,
+ * find a continuous sub-array which adds to a given number S, If not found
+print -1.
+ *
+ * Input:
```
kvedala, line 9: Please elaborate on the input. What is the first line with only one value? What is the second line with two values?

others/sub_array_with_given_sum.cpp
```diff
@@ -0,0 +1,89 @@
...
+void test() {
+    std::cout << "Test 1" << std::endl;
+    int arr1[] = {1, 2, 3, 7, 5};
+    int original_sum = 12;
+    int size = sizeof(arr1) / sizeof(arr1[0]);
+    contiguous_sub_array(arr1, size, original_sum);
+    std::cout << "Test 1 passed\n ===================== \n";
+
+    std::cout << "Test 2" << std::endl;
+    int arr2[] = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
```
kvedala, line 78: Would be a good idea to test with random arrays

others/sub_array_with_given_sum.cpp
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Function to print the index range of subarray equal to the given sum. If no
+ * such subarray found, print -1.
+ * @param arr array to check
+ * @param size size of an array
+ * @param original_sum sum to match with the sum of all elements of a sub array.
+ * @return void and prints index ranges if sum of sub array equal to the given
+ * sum, else prints -1.
+ */
+void contiguous_sub_array(int *arr, int size, int orignal_sum) {
```
kvedala, line 38: The sum should be returned as function return value. Printing it should be done by the calling function.

others/sub_array_with_given_sum.cpp
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Test cases to test the program
+ */
+void test() {
+    std::cout << "Test 1" << std::endl;
+    int arr1[] = {1, 2, 3, 7, 5};
+    int original_sum = 12;
+    int size = sizeof(arr1) / sizeof(arr1[0]);
+    contiguous_sub_array(arr1, size, original_sum);
+    std::cout << "Test 1 passed\n ===================== \n";
```
kvedala, line 75: There is no check being performed. Self checks use `assert` statements to charitable results. Please check existing code for references.

others/sub_array_with_given_sum.cpp
omkarlanghe: Change done as per the above review comment.

others/sub_array_with_given_sum.cpp
omkarlanghe: I have added assert. It checks for an expression assert(ivec.size() > 1). If -1 encountered, it throws error.

kvedala: this would check if there was a nonzero length of substring found.
What if the substring doesn't exist that satisfies the condition?

The check should be for expected substring to be returned. If the substring in a test case is an array containing `{3,4,5}`, then these are the numbers to be checked. Something like:
```cpp
// this is the expected output
std::vector<int> expected_substring({1,2,3});

std::vector<int> result; // store output from the function

// invoke function
contiguous_sub_array(....); // this function should update the `result` variable

assert(result == expected_substring); // this now checks if the result is expected substring i.e., did the function perform correctly or not
```
Moreover, the check should be performed *outside* of the function you want to test.

PR #884, opened 2020-06-23
Feat: Class implementation of complex numbers along with most complex number operations.

tjgurwara99:
#### Description of Change

A class implementation of complex numbers along with all permitted field operations and some additional complex number specific operations.

#### Checklist

Notes:  Class implementation of complex numbers.

### DIFF
--- a/math/complex_numbers.cpp
+++ b/math/complex_numbers.cpp
@@ -2,11 +2,12 @@
  * Copyright 2020 @author tjgurwara99
  * @file
  *
- * A basic implementation of Complex Number field as a class with operators overloaded to accommodate (mathematical) field operations.
+ * A basic implementation of Complex Number field as a class with operators
+ * overloaded to accommodate (mathematical) field operations.
  */
 
-#include <iostream>
 #include <cmath>
+#include <iostream>
 #include <stdexcept>
 
 /**
@@ -18,9 +19,10 @@ class Complex {
     // The imaginary value of the complex number
     double im;
 
- public:
+  public:
     /**
-     * Complex Constructor which initialises the complex number which takes two arguments.
+     * Complex Constructor which initialises the complex number which takes two
+     * arguments.
      * @param x The real value of the complex number.
      * @param y The imaginary value of the complex number.
      */
@@ -28,34 +30,31 @@ class Complex {
         this->re = x;
         this->im = y;
     }
-    
+
     /**
-     * Complex Constructor which initialises the complex number with no arguments.
+     * Complex Constructor which initialises the complex number with no
+     * arguments.
      */
-    Complex() {
-        Complex(0.0,0.0);
-    }
+    Complex() { Complex(0.0, 0.0); }
 
     /**
      * Member function (getter) to access the class' re value.
      */
-    double real() const {
-        return this->re;
-    }
+    double real() const { return this->re; }
 
     /**
      * Member function (getter) to access the class' im value.
      */
-    double imag() const {
-        return this->im;
-    }
+    double imag() const { return this->im; }
 
     /**
-     * Member function to which gives the absolute value (modulus) of our complex number
-     * @return \f$ \sqrt{z \dot \bar{z} \f$ where \f$ z \f$ is our complex number.
+     * Member function to which gives the absolute value (modulus) of our
+     * complex number
+     * @return \f$ \sqrt{z \dot \bar{z}} \f$ where \f$ z \f$ is our complex
+     * number.
      */
     double abs() const {
-        return std::sqrt(this->re*this->re + this->im*this->im);
+        return std::sqrt(this->re * this->re + this->im * this->im);
     }
 
     /**
@@ -63,7 +62,7 @@ class Complex {
      * @param other The other number that is added to the current number.
      * @return result current number plus other number
      */
-    Complex operator+(const Complex& other) {
+    Complex operator+(const Complex &other) {
         Complex result(this->re + other.re, this->im + other.im);
         return result;
     }
@@ -73,7 +72,7 @@ class Complex {
      * @param other The other number being subtracted from the current number.
      * @return result current number subtract other number
      */
-    Complex operator-(const Complex& other) {
+    Complex operator-(const Complex &other) {
         Complex result(this->re - other.re, this->im - other.im);
         return result;
     }
@@ -83,15 +82,16 @@ class Complex {
      * @param other The other number to multiply the current number to.
      * @return result current number times other number.
      */
-    Complex operator*(const Complex& other) {
+    Complex operator*(const Complex &other) {
         Complex result(this->re * other.re - this->im * other.im,
                        this->re * other.im + this->im * other.re);
         return result;
     }
 
     /**
-     * Operator overload of the BITWISE NOT which gives us the conjugate of our complex number.
-     * NOTE: This is overloading the BITWISE operator but its not a BITWISE operation in this definition.
+     * Operator overload of the BITWISE NOT which gives us the conjugate of our
+     * complex number. NOTE: This is overloading the BITWISE operator but its
+     * not a BITWISE operation in this definition.
      * @return result The conjugate of our complex number.
      */
     Complex operator~() const {
@@ -100,18 +100,19 @@ class Complex {
     }
 
     /**
-     * Operator overload to be able to divide two complex numbers. This function would throw an exception if the other number is zero.
+     * Operator overload to be able to divide two complex numbers. This function
+     * would throw an exception if the other number is zero.
      * @param other The other number we divide our number by.
      * @return result Current number divided by other number.
      */
-    Complex operator/(const Complex& other) {
+    Complex operator/(const Complex &other) {
         Complex result = *this * ~other;
         double denominator = other.abs() * other.abs();
         if (denominator != 0) {
-            result = Complex(result.real() / denominator, result.imag() / denominator);
+            result = Complex(result.real() / denominator,
+                             result.imag() / denominator);
             return result;
-        }
-        else {
+        } else {
             throw std::invalid_argument("Undefined Value");
         }
     }
@@ -124,23 +125,24 @@ class Complex {
  * @return 'True' If real and imaginary parts of a and b are same
  * @return 'False' Otherwise.
  */
-bool operator==(const Complex& a, const Complex& b) {
+bool operator==(const Complex &a, const Complex &b) {
     double del_real = a.real() - b.real();
     double del_imag = a.imag() - b.imag();
-    return ((del_real <= 1e-15 && del_real >= - 1e-15 ) && (del_imag <= 1e-15 && del_imag >= - 1e-15));
+    return ((del_real <= 1e-15 && del_real >= -1e-15) &&
+            (del_imag <= 1e-15 && del_imag >= -1e-15));
 }
 
 /**
- * Overloaded insersion operator to accommodate the printing of our complex number in their standard form.
+ * Overloaded insersion operator to accommodate the printing of our complex
+ * number in their standard form.
  * @param os The console stream
  * @param num The complex number.
  */
-std::ostream& operator<<(std::ostream& os, const Complex& num) {
+std::ostream &operator<<(std::ostream &os, const Complex &num) {
     os << num.real();
     if (num.imag() < 0) {
         os << " - " << -num.imag();
-    }
-    else {
+    } else {
         os << " + " << num.imag();
     }
     os << "i";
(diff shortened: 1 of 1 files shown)


math/complex_numbers.cpp
```diff
@@ -0,0 +1,190 @@
...
+    double im;
+
+  public:
+    /**
+     * Complex Constructor which initialises the complex number which takes two
+     * arguments.
+     * @param x The real value of the complex number.
+     * @param y The imaginary value of the complex number.
+     */
+    Complex(double x, double y) {
```
kvedala, line 30: C++11 allows us to instantiate with default values. Also, `this` keyword incorrect as the instance is not complete yet. The correct way is:
```suggestion
    Complex(double x 0 0.f, double y = 0.f) : re(x), im(y) {;
```

math/complex_numbers.cpp
```diff
@@ -0,0 +1,190 @@
...
+        this->re = x;
```
kvedala, line 31:
```suggestion
```

math/complex_numbers.cpp
```diff
@@ -0,0 +1,190 @@
...
+        this->im = y;
```
kvedala, line 32:
```suggestion
```

math/complex_numbers.cpp
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * Complex Constructor which initialises the complex number with no
+     * arguments.
+     */
+    Complex() { Complex(0.0, 0.0); }
```
kvedala:
```suggestion
```

kvedala: Needs a *copy* or *assignment* operator:
```cpp
Complex(const Complex &other) : re(other.real), im(other.im) {;}
```

math/complex_numbers.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Tests Function
+ */
+void tests() {
+    Complex num1(1, 1), num2(1, 1);
+    // Test for addition
+    assert(((void)"1 + 1i + 1 + 1i is equal to 2 + 2i but the addition doesn't "
+                  "add up \n",
+            (num1 + num2) == Complex(2, 2)));
```
kvedala, line 164: I think the correct way to check would be to compare with the results from `std::complex` class.

math/complex_numbers.cpp
```diff
@@ -0,0 +1,190 @@
...
+     */
+    double imag() const { return this->im; }
+
+    /**
+     * Member function to which gives the absolute value (modulus) of our
+     * complex number
+     * @return \f$ \sqrt{z \dot \bar{z}} \f$ where \f$ z \f$ is our complex
+     * number.
+     */
+    double abs() const {
```
kvedala, line 60: please include `double arg() const {` for the argument value of a complex number

math/complex_numbers.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+/**
+ * Overloaded insersion operator to accommodate the printing of our complex
+ * number in their standard form.
+ * @param os The console stream
+ * @param num The complex number.
+ */
+std::ostream &operator<<(std::ostream &os, const Complex &num) {
+    os << num.real();
```
kvedala, line 146: maybe put the complex number within parentheses?

math/complex_numbers.cpp
tjgurwara99: Oh yeah, I forgot about that😅

math/complex_numbers.cpp
tjgurwara99: Sure, will add that now.

math/complex_numbers.cpp
tjgurwara99: All tests work except for the one that checks for division.

math/complex_numbers.cpp
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * Operator overload to be able to divide two complex numbers. This function
+     * would throw an exception if the other number is zero.
+     * @param other The other number we divide our number by.
+     * @return result Current number divided by other number.
+     */
+    Complex operator/(const Complex &other) {
+        Complex result = *this * ~other;
+        double denominator = other.abs() * other.abs();
+        if (denominator != 0) {
```
kvedala, line 115: 1. when using doubles, check for `std::numeric_limits<double>::epsilon()` instead of '0'. This is defined in `<limits>` header.
2. when checking the result with output from `std::complex` class, do not check for equality but check for approximately equal to. refer [this example code](https://github.com/TheAlgorithms/C-Plus-Plus/blob/27a63f54275f14b68e96b1ddea075fd4c29e4f51/numerical_methods/brent_method_extrema.cpp#L178)

math/complex_numbers.cpp
tjgurwara99: I changed my implementation to coincide with the standard libraries.

math/complex_numbers.cpp
tjgurwara99: I tweaked my implementations to coincide with the standard libraries to make sure they are equal but I can change it to work with epsilon as well if that is what's preferred😄. Let me know

tjgurwara99: Also which c++ standard is being used in CI workflow? There are some very cool features in C++20 like [this](https://stackoverflow.com/a/57285400)

kvedala: We are currently using C++11 only because if people want to use on older compilers or on embedded platforms like a microcontroller, they still can. Most high-level additions in post C++11 are not supported by the ARM cortex or Tensilica, etc processors.

Also, this is something I suggest but feel free to bring it up in the feature request/issues discussions.

math/complex_numbers.cpp
```diff
@@ -0,0 +1,238 @@
...
+    }
+    os << "i)";
+    return os;
+}
+
+/**
+ * Tests Function
+ */
+void tests() {
+    Complex num1(1, 1), num2(1, 1);
```
kvedala, line 188: would it be better to replace the hard-coded numbers by something like:
```suggestion
    double get_rand(){
       return (std::rand() % 100 - 50) / 100.f;
    }

    std::srand(std::time(nullptr));
    Complex num1(get_rand(), get_rand()), num2(get_rand(), get_rand());
```
What are the pros and cons?

math/complex_numbers.cpp
```diff
@@ -0,0 +1,238 @@
...
+    std::complex<double> cnum1(1, 1), cnum2(1, 1);
+    // Test for addition
+    assert(((void)"1 + 1i + 1 + 1i is equal to 2 + 2i but the addition doesn't "
+                  "add up \n",
+            ((num1 + num2).real() == (cnum1 + cnum2).real() &&
+             (num1 + num2).imag() == (cnum1 + cnum2).imag())));
+    std::cout << "First test passes." << std::endl;
+    // Test for subtraction
+    assert(((void)"1 + 1i - 1 - 1i is equal to 0 but the program says "
```
kvedala, line 197: better to do the operation once and check the results
```cpp
Complex result = num1 - num2;
std::complex<double> expected = cnum1 - cnum2;
assert(result.real() == expected.real && result.imag() == expected.imag());
```

math/complex_numbers.cpp
```diff
@@ -0,0 +1,238 @@
...
+#include <cassert>
+#include <cmath>
+#include <complex>
+#include <iostream>
+#include <stdexcept>
+
+/**
+ * Global constant for pi to work with Windows environment
+ */
+const double pi = 3.14159265358979323846;
```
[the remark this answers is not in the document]
tjgurwara99: Ok, should I change it? I used `M_PI` but not the `#define _USE_MATH_DEFINES`  so probably that's why it didn't work in Windows CI workflow.

math/complex_numbers.cpp
```diff
@@ -0,0 +1,238 @@
...
+    double abs() const {
+        return std::sqrt(this->re * this->re + this->im * this->im);
+    }
+
+    /**
+     * Member function which gives the argument of our complex number.
+     * @return Argument of our Complex number.
+     */
+    double arg() const {
+        if (this->re > 0) {
```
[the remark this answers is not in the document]
tjgurwara99: Oh I didn't know about this one

math/complex_numbers.cpp
```diff
@@ -0,0 +1,238 @@
...
+     * @return \f$ \sqrt{z \dot \bar{z}} \f$ where \f$ z \f$ is our complex
+     * number.
+     */
+    double abs() const {
+        return std::sqrt(this->re * this->re + this->im * this->im);
+    }
+
+    /**
+     * Member function which gives the argument of our complex number.
+     * @return Argument of our Complex number.
```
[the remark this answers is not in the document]
tjgurwara99: Radians. I'll add it in the documentation comment.

math/complex_numbers.cpp
tjgurwara99: None that I can foresee.

math/complex_numbers.cpp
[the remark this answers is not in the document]
kvedala: > Ok, should I change it? I used `M_PI` but not the `#define _USE_MATH_DEFINES` so probably that's why it didn't work in Windows CI workflow.

Correct. In the references I shared with you, I have that specifically documented on the very first line.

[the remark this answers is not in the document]
tjgurwara99: Ok so using atan2 makes this unnecessary so I removed it.

math/complex_numbers.cpp
```diff
@@ -0,0 +1,238 @@
+/**
+ * Copyright 2020 @author tjgurwara99
```
kvedala, line 2:
```suggestion
 * @author tjgurwara99
```
No need to add "copyright message". The repo already has an associated license.

math/complex_numbers.cpp
```diff
@@ -63,23 +58,9 @@ class Complex {
...
-            return std::atan(this->im / this->re) - pi;
-        } else if (this->re == 0 && this->im > 0) {
-            return pi / 2;
-        } else if (this->re == 0 && this->im < 0) {
-            return -pi / 2;
-        } else {
-            throw std::invalid_argument("Undefined Value");
-        }
-    }
+    double arg() const { return atan2(this->im, this->re); }
```
kvedala, line 64:
```suggestion
    double arg() const { return std::atan2(this->im, this->re); }
```

kvedala:
Should there be an instantiation of the complex number from a given magnitude and angle?

tjgurwara99:
> Should there be an instantiation of the complex number from a given magnitude and angle?

I was thinking of that but I don't know how I should go about overloading the Constructor with two doubles because we would need to use two doubles for abs and arg. Any recommendations as to what we should use?

kvedala:
> > Should there be an instantiation of the complex number from a given magnitude and angle?
> 
> I was thinking of that but I don't know how I should go about overloading the Constructor with two doubles because we would need to use two doubles for abs and arg. Any recommendations as to what we should use? @kvedala

ah, good point. how about 
```cpp
Complex(double x=0.f, double y=0.f, bool is_polar=false)
```
if `is_polar` is set to true, assume that x is magnitude and y is angle. But the documentation needs to be clear

tjgurwara99:
> > > Should there be an instantiation of the complex number from a given magnitude and angle?
> > 
> > 
> > I was thinking of that but I don't know how I should go about overloading the Constructor with two doubles because we would need to use two doubles for abs and arg. Any recommendations as to what we should use? @kvedala
> 
> ah, good point. how about
> 
> ```c++
> Complex(double x=0.f, double y=0.f, bool is_polar=false)
> ```
> 
> if `is_polar` is set to true, assume that x is magnitude and y is angle. But the documentation needs to be clear

I've added this feature but I'm not sure about the readability of the coding style. Can you take a look? @kvedala

math/complex_numbers.cpp
```diff
@@ -0,0 +1,249 @@
...
+     * x is the real value of the complex number (optional).
+     * @param y If the third parameter is 'true' then this y is the argument of
+     * the complex number, if the third parameter is 'false' then this y is the
+     * imaginary value of the complex number (optional).
+     * @param is_polar 'false' by default. If we want to initialise our complex
+     * number using polar form then set this to true, otherwise set it to false
+     * to use initialiser which initialises real and imaginary values using the
+     * first two parameters (optional).
+     */
+    explicit Complex(double x = 0.f, double y = 0.f, bool is_polar = false)
```
kvedala, line 40: agreed about the readability.
```suggestion
    explicit Complex(double x = 0.f, double y = 0.f, bool is_polar = false) {
       if (!polar) {
         re = x;
         im = y;
         return;
       }
       
       re = x * std::cos(y);
       im = x * std::sin(y);
   }
```

kvedala:
@tjgurwara99 you can convert this to a C file and pull it to: https://github.com/TheAlgorithms/C

tjgurwara99:
> lgtm

Is there a way to check the lgtm response quickly on my local pc without pull requests? @kvedala 

> @tjgurwara99 you can convert this to a C file and pull it to: https://github.com/TheAlgorithms/C

Also, I'll write this soon😄. Thanks!

kvedala:
> > lgtm
> 
> Is there a way to check the lgtm response quickly on my local pc without pull requests? @kvedala
> 
> > @tjgurwara99 you can convert this to a C file and pull it to: https://github.com/TheAlgorithms/C
> 
> Also, I'll write this soon😄. Thanks!

It seems that LGTM is simply taking the checks made by `clang-tidy` and making a pretty presentation. From what I understand, to use LGTM features locally would require additional coding on your part. But if you are ok, then you can use the `clang-tidy` command line utility. I have an implementation in a branch on my fork - but it is very strict in its checks and much of existing code is still not unto the standard and it fails there. Feel free to play with it - https://github.com/kvedala/C-Plus-Plus/tree/fix_ci and give any feedback. If you do end up improving it, do share 😄

kvedala:
> > > lgtm
> > 
> > 
> > Is there a way to check the lgtm response quickly on my local pc without pull requests? @kvedala
> > > @tjgurwara99 you can convert this to a C file and pull it to: https://github.com/TheAlgorithms/C
> > 
> > 
> > Also, I'll write this soon😄. Thanks!
> 
> It seems that LGTM is simply taking the checks made by `clang-tidy` and making a pretty presentation. From what I understand, to use LGTM features locally would require additional coding on your part. But if you are ok, then you can use the `clang-tidy` command line utility. I have an implementation in a branch on my fork - but it is very strict in its checks and much of existing code is still not unto the standard and it fails there. Feel free to play with it - https://github.com/kvedala/C-Plus-Plus/tree/fix_ci and give any feedback. If you do end up improving it, do share 😄

The branch with clang-tidy got deleted during one of the previous merges. but in the [GitHub Actions](https://github.com/kvedala/C-Plus-Plus/blob/f956c446f10891a16068232719326d366fc75254/.github/workflows/awesome_workflow.yml#L48] replace `clang-format`, with `clang-tidy` specifications.

kvedala:
See the generated output - https://thealgorithms.github.io/C-Plus-Plus/da/d5a/class_complex.html and https://thealgorithms.github.io/C-Plus-Plus/d5/d67/complex__numbers_8cpp.html
Notice the following:
1. the file does not have a `\brief` section in the beginning. Hence, in the list of files table, there is no description for this file
2. The math syntax is incorrect in the description of function `abs`

tjgurwara99:
> 1. the file does not have a `\brief` section in the beginning. Hence, in the list of files table, there is no description for this file

I'll create a new PR after adding `\brief` tag.

> 2. The math syntax is incorrect in the description of function `abs`

It seems that the inline latex is not recognising the `\bar` command. I can try to replace it with `\overline` but I think if `\bar` didn't work then `\overline` is also unlikely to work. The raw comment is this `@return \f$ \sqrt{z \dot \bar{z}} \f$ where ...`. Any idea why that is the case?

tjgurwara99:
After looking over the latex in a latex editor it does seem that the syntax would is not correct for what it is supposed to show.

I'll change it in the PR to see if it changes the output. Thanks for noticing this @kvedala!

PR #891, opened 2020-06-24
docs: add comments and explanation to binary_search_tree.cpp data_structures

liushubin-gitHub:
#### Description of Change
This pull request make the following changes:
1. docs: add comments and explanation to binary_search_tree.cpp data_structures

2. test: add test for binary_search_tree.cpp data_structures

3. fix: Remove() function bugs in binary_search_tree.cpp data_structures

#### Checklist

- [x ] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#Commit-Guidelines)
- [x ] Search previous suggestions before making a new one, as yours may be a duplicate.
- [x ] I acknowledge that all my contributions will be made under the project's license.

@kvedala Pls, help review!
Notes: 

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/891"></a>

### DIFF
--- a/data_structures/binary_search_tree.cpp
+++ b/data_structures/binary_search_tree.cpp
@@ -7,7 +7,8 @@
  * utilize any of the C++ STL features.
  */
 #include <iostream>
-
+#include <algorithm>//test code
+#include <cassert>
 /** An struct node type.
  *  Define a element for tree insert/remove operation
  */
@@ -155,6 +156,17 @@ void Post(node *n) {
     }
 }
 
+int arr[7] = { 0 };//for test tree use only
+int index = 0;//for test tree use only
+void testInOrderTraverse(node* n) {
+    if (n != NULL) {
+        testInOrderTraverse(n->left);
+        std::cout << n->val << "  ";
+        arr[index++] = n->val;
+        testInOrderTraverse(n->right);
+    }
+}
+
 void test_tree() {
     node *root = new node;
     root->val = 4;
@@ -170,13 +182,22 @@ void test_tree() {
     std::cout
         << "after Insert() ,the expected output should be : 1, 2, 3, 4, 5, 6 ,7"
         << std::endl;
-    In(root);
+    //test code
+    testInOrderTraverse(root);
+    assert(std::is_sorted(arr, arr + 7));
+    std::cout << "Test Insert() function Passed\n========================\n";
+
     // test Remove()
+    index = 0;
     Remove(root, root, 2);
     std::cout << "\n after Remove() node 2 , the expected output should be : "
                  "1, 3, 4, 5, 6, 7"
               << std::endl;
-    In(root);
+    testInOrderTraverse(root);
+    for (int i = 0; i < 6; i++) {
+        assert(arr[i] != 2);
+    }
+    std::cout << "Test Remove(2) Passed\n========================\n";
 }
 
 int main() {
--- a/data_structures/binary_search_tree.cpp
+++ b/data_structures/binary_search_tree.cpp
@@ -9,15 +9,17 @@
 #include <algorithm>  // test code
 #include <cassert>
 #include <iostream>
+#include <memory>
 #include <queue>
 #include <vector>
-/** An struct node type.
- *  Define a element for tree insert/remove operation
- */
+
+ /** An struct node type.
+  *  Define a element for tree insert/remove operation
+  */
 struct node {
-    int val;     /**< int value of a node struct*/
-    node *left;  /**< left subtree pointer */
-    node *right; /**< right subtree pointer */
+    int val{};                   /**< int value of a node struct*/
+    std::shared_ptr<node> left;  /**< left subtree pointer */
+    std::shared_ptr<node> right; /**< right subtree pointer */
 };
 
 /** insert a node to tree
@@ -25,27 +27,27 @@ struct node {
  * \param[in] n root node of a tree
  * \param[in] x a node with value to be insert
  */
-void Insert(node *n, int x) {
-    node *temp = nullptr;
+void Insert(node* n, int x) {
     if (x < n->val) {
         if (n->left == nullptr) {
-            temp = new node;
-            temp->val = x;
-            temp->left = nullptr;
-            temp->right = nullptr;
-            n->left = temp;
-        } else {
-            Insert(n->left, x);
+            n->left = std::shared_ptr<node>(new node);
+            n->left->val = x;
+            n->left->left = nullptr;
+            n->left->right = nullptr;
         }
-    } else {
+        else {
+            Insert(n->left.get(), x);
+        }
+    }
+    else {
         if (n->right == nullptr) {
-            temp = new node;
-            temp->val = x;
-            temp->left = nullptr;
-            temp->right = nullptr;
-            n->right = temp;
-        } else {
-            Insert(n->right, x);
+            n->right = std::shared_ptr<node>(new node);
+            n->right->val = x;
+            n->right->left = nullptr;
+            n->right->right = nullptr;
+        }
+        else {
+            Insert(n->right.get(), x);
         }
     }
 }
@@ -54,9 +56,9 @@ void Insert(node *n, int x) {
  * \param[in] n the root node pointer of subtree
  * \return the max node int value
  */
-int findMaxInLeftST(node *n) {
+int findMaxInLeftST(node* n) {
     while (n->right != nullptr) {
-        n = n->right;
+        n = n->right.get();
     }
     return n->val;
 }
@@ -67,37 +69,48 @@ int findMaxInLeftST(node *n) {
  * \param[in] n start node to search a node with value x
  * \param[in] x the int value of a node
  */
-void Remove(node *p, node *n, int x) {
+void Remove(std::shared_ptr<node> p, std::shared_ptr<node> n, int x) {
     if (n->val == x) {
         if (n->right == nullptr && n->left == nullptr) {
             if (x > p->val) {
                 p->right = nullptr;
-            } else {
+            }
+            else {
                 p->left = nullptr;
             }
-            delete n;
-        } else if (n->right == nullptr) {
+            //n.reset(nullptr);
+            //   delete n;
+        }
+        else if (n->right == nullptr) {
             if (x > p->val) {
                 p->right = n->left;
-            } else {
+            }
+            else {
                 p->left = n->left;
             }
-            delete n;
-        } else if (n->left == nullptr) {
+            //n.reset(nullptr);
+            //   delete n;
+        }
+        else if (n->left == nullptr) {
             if (x > p->val) {
                 p->right = n->right;
-            } else {
+            }
+            else {
                 p->left = n->right;
             }
-            delete n;
-        } else {
-            int y = findMaxInLeftST(n->left);
+            //n.reset(nullptr);
+            //   delete n;
+        }
+        else {
+            int y = findMaxInLeftST(n->left.get());
             n->val = y;
             Remove(n, n->left, y);
         }
-    } else if (x < n->val) {
+    }
+    else if (x < n->val) {
         Remove(n, n->left, x);
-    } else {
+    }
+    else {
         Remove(n, n->right, x);
     }
 }
@@ -105,12 +118,13 @@ void Remove(node *p, node *n, int x) {
 /** free memory of all tree nodes.
  * \param[in] n the root node pointer of a tree
  */
-void FreeTreeNodes(node *root) {
+void FreeTreeNodes(std::shared_ptr<node> root) {
     if (root != nullptr) {
-        FreeTreeNodes(root->left);
-        FreeTreeNodes(root->right);
+        FreeTreeNodes(std::move(root->left));
+        FreeTreeNodes(std::move(root->right));
         std::cout << root->val << "  ";
-        delete root;  // finaly delete root node
+        //root.reset(nullptr);
+        // delete root; // finaly delete root node
     }
 }
 
@@ -118,21 +132,23 @@ void FreeTreeNodes(node *root) {
  * and begin with node pointer n
  * \param[in] n node pointer of a tree
  */
-void BFT(node *n) {
-    std::queue<node *> queue;
+void BFT(node* n) {
+    std::queue<node*> queue;
     if (n != nullptr) {
         queue.push(n);
     }
 
-    node *temp = nullptr;
+    node* temp = nullptr;
     while (!queue.empty()) {
         temp = queue.front();
         queue.pop();
         std::cout << temp->val << "  ";
-        if (temp->left)
-            queue.push(temp->left);
-        if (temp->right)
-            queue.push(temp->right);
+        if (temp->left) {
+            queue.push(temp->left.get());
+        }
+        if (temp->right) {
+            queue.push(temp->right.get());
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


data_structures/binary_search_tree.cpp
```diff
@@ -122,7 +155,33 @@ void Post(node *n) {
...
+    root->left = NULL;
+    root->right = NULL;
+    // test Insert()
+    Insert(root, 2);
+    Insert(root, 1);
+    Insert(root, 3);
+    Insert(root, 6);
+    Insert(root, 5);
+    Insert(root, 7);
+    std::cout
```
kvedala, line 170: should perform self-check without manual intervention; something like below:
```cpp
for(int i = 1; i < 8; i++)
    assert(root[i] == i);
```

kvedala: Also, `2` is inserted first, so, shouldn't the expected output be 
`2,1,3,6,5,7`?

liushubin-gitHub: > should perform self-check without manual intervention; something like below:
> 
> ```c++
> for(int i = 1; i < 8; i++)
>     assert(root[i] == i);
> ```

OK，I have use assert() instead. 
Many thanks!

liushubin-gitHub: > Also, `2` is inserted first, so, is the output expected to be
> `2,1,3,6,5,7` or
> `1, 2, 3, 4, 5, 6 ,7`?
Here, i want to build a special tree.
When use Inorder traversal a binary search tree, the ouput is a sorted array:)

kvedala:
rule of thumb - never use global variables

liushubin-gitHub:
I will find another way to make it better.

liushubin-gitHub:
fix lgtm alerts: remove global variables.

kvedala:
@liushubin-gitHub any update on this with fixing the errors? Thank you

liushubin-gitHub:
yes it is done .
lgtm  has passed

kvedala:
> yes it is done .
> lgtm has passed

Thank you.
There are still undocumented structures and functions. Can you please document them as well.

liushubin-gitHub:
> > yes it is done .
> > lgtm has passed
> 
> Thank you.
> There are still undocumented structures and functions. Can you please document them as well.

yes，I will try my best to do it.
But there is still some functions that i don't understand clearly,so i just leave it empty. (as i am not the original author)
Also, there should be some  document about test cases  , but  i  think not so much important (test code is clearly enough), then i just skip :)

kvedala:
> > > yes it is done .
> > > lgtm has passed
> > 
> > 
> > Thank you.
> > There are still undocumented structures and functions. Can you please document them as well.
> 
> yes，I will try my best to do it.
> But there is still some functions that i don't understand clearly,so i just leave it empty. (as i am not the original author)
> Also, there should be some document about test cases , but i think not so much important (test code is clearly enough), then i just skip :)

Yes, there are some codes that are not properly documented. It is an ongoing effort to fic them: see for example #554 #555

liushubin-gitHub:
> > > > yes it is done .
> > > > lgtm has passed
> > > 
> > > 
> > > Thank you.
> > > There are still undocumented structures and functions. Can you please document them as well.
> > 
> > 
> > yes，I will try my best to do it.
> > But there is still some functions that i don't understand clearly,so i just leave it empty. (as i am not the original author)
> > Also, there should be some document about test cases , but i think not so much important (test code is clearly enough), then i just skip :)
> 
> Yes, there are some codes that are not properly documented. It is an ongoing effort to fic them: see for example #554 #555
> Leaving code half-fixed will lead to issues in maintaining them as they will be inconsistent and contributors can have the wrong idea. Moreover, the repo auto-generates a documentation website - https://thealgorithms.github.io/C that will not render correctly if code is only half fixed.

Got it，i will try it again later

liushubin-gitHub:
Pls review!

kvedala:
Please fix the errors reported by the system. You can click on "details" to see the error details

data_structures/binary_search_tree.cpp
```diff
@@ -122,9 +159,64 @@ void Post(node *n) {
...
+        testInOrderTraverse(n->right, arr);
+    }
+}
+
+/** test function.
+ * auto build a binary tree with special nodes.
+ * test the tree Insert() and Remove() function and out test result.
+ */
+void test_tree() {
+    node *root = new node;
```
kvedala, line 181: Dynamic memory allocated here must be freed.

liushubin-gitHub: yes，fixed

kvedala: Though the applied fix works, can you implement a function to which, a parent node can be passed and it will iteratively free its children? If not, this is still acceptable.

kvedala:
image
please check the errors reported

liushubin-gitHub:
> image
> please check the errors reported

i found the code format fail has been auto corrected by the code formatter!

kvedala:
> > image
> > please check the errors reported
> 
> i found the code format fail has been auto corrected by the code formatter!

I don't think so. Can you show me where you see that? 
Did you actually click the `details` link and see the error message?

kvedala:
Screen Shot 2020-07-10 at 2 01 31 PM
for you, I have taken time to open the link, mark the line that describes the error.

liushubin-gitHub:
> > > image
> > > please check the errors reported
> > 
> > 
> > i found the code format fail has been auto corrected by the code formatter!
> 
> I don't think so. Can you show me where you see that?
> Did you actually click the `details` link and see the error message?
i am try to find the details .  there is an auto commit of system:
https://github.com/TheAlgorithms/C-Plus-Plus/pull/891/commits/9dafdfa15f31bbd3cd69c977dba0b570773dcb4a

data_structures/binary_search_tree.cpp
```diff
@@ -122,9 +159,72 @@ void Post(node *n) {
     }
 }
 
+/** test traverse a tree with Post order
+ * and save the traverse node to an array.
+ * \param[in] n the node pointer of a tree to start traverse
+ * \param[out] arr the output array that save the traversed tree nodes pointers
+ */
+void testInOrderTraverse(node *n, std::vector<int> &arr) {
```
kvedala, line 167:
```
void testInOrderTraverse(node *n, std::vector<int> &arr) {
```
This line is the error. the error describes that `arr` must be a const reference. If it is not, use a pointer.
```suggestion
void testInOrderTraverse(node *n, std::vector<int> *arr) {
```

liushubin-gitHub:
> Screen Shot 2020-07-10 at 2 01 31 PM
> for you, I have taken time to open the link, mark the line that describes the error.

Thanks you very much. 
I will check autotest erros more carefully next time.

data_structures/binary_search_tree.cpp
liushubin-gitHub: Good idea! I  have fixed it.

kvedala:
I have marked this particular type of error (https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-owning-memory.html) as a no-error but warning.
Further commits will not mark this as error.

kvedala:
The correct way to resolve them is by using `std::unique_ptr<node*> temp = new node` instead of `node* temp = new node;`

liushubin-gitHub:
> unique_ptr

I will try it once more.

kvedala:
> > unique_ptr
> 
> I will try it once more.

I have made all the fixes for `unique_ptr` for you and created a PR to your fork. https://github.com/liushubin-gitHub/C-Plus-Plus/pull/4

liushubin-gitHub:
image
how to solve the auto test error? 
could it merged successfully now?

liushubin-gitHub:
image

the auto test passed ?

kvedala:
> image
> how to solve the auto test error?
> could it merged successfully now?

Fixed. Another commit should be able to check cirrectly

kvedala:
> image
> 
> the auto test passed ?

No, this is only one check and after it did code formatting. You have to check the commit by you, not by github-actions

kvedala:
now it looks fixed

kvedala:
Your last commit broke the code. 
Smart pointers delete the memory when the last reference to the memory is deleted.

liushubin-gitHub:
https://github.com/TheAlgorithms/C-Plus-Plus/pull/891#issuecomment-657284669

now everything is fine.
auto test ok.
execution test ok.

kvedala:
it still gives me a segmentation fault:
Screen Shot 2020-07-14 at 8 01 20 AM
I executed it on GitPod using the link in the Pull-request description.

kvedala:
> it still gives me a segmentation fault:
> Screen Shot 2020-07-14 at 8 01 20 AM
> I executed it on GitPod using the link in the Pull-request description.

the function `Remove()`, probably needs a `null` check?

data_structures/binary_search_tree.cpp
```diff
@@ -75,6 +76,10 @@ int findMaxInLeftST(node* n) {
  * \param[in] x the int value of a node
  */
 void Remove(std::shared_ptr<node>* p, std::shared_ptr<node>* n, int x) {
+    if (*n == nullptr) {
```
kvedala, line 79: What you had originally was also fine.
you can pass `p` and `n` and `const std:shared_ptr<node>` (this is the appropriate way) and still do the 

```cpp
if (n == nullptr) return;
```

liushubin-gitHub: yes, I try to used const reference as you suggested , but const reference dosen't work well here.
We need to modify the "node n" reference value  in function Remove().
So , the system suggest  to use pointer.

kvedala:
👍 
Can you add the case : any number not in the menu list should exit the program?

liushubin-gitHub:
Pls help review!

kvedala (dismissed):
👍 
@ayaankhan98  please review as well. thanks.

kvedala:
You you add that in the `std::out` statements that any other number will exit?

data_structures/binary_search_tree.cpp
```diff
@@ -1,138 +1,246 @@
...
-node *dequeue() { return (queue.t[queue.front++]); }
-
-void Insert(node *n, int x) {
+/** insert a node to tree
+ * with greater value to right subtree, smaller value node to left subtree
+ * \param[in] n root node of a tree
+ * \param[in] x a node with value to be insert
+ */
+void Insert(node* n, int x) {
+    assert(n != NULL);
```
ayaankhan98, line 39: what happens if the user inserts the very first root element? this Insert function will break in that case? you are not handling the root insertition case, instead you assumed to insert the root value manually as mentioned in the test function
```
    root->val = 4;
    root->left = nullptr;
    root->right = nullptr;
```
i suggest you to handle this too in the insert function because basically we are performing the very same operation that is we are insterting a value let it be anywhere in the tree either the root or the child everything should be properly handled by the same insert function

liushubin-gitHub:
> any other number will exit

done

data_structures/binary_search_tree.cpp
liushubin-gitHub: @ayaankhan98 Thanks for you suggestion. I have got your means: 
you want to design a function that is very common to use.
But , since here we are talking about a special tree (binary search tree).

ayaankhan98: https://github.com/ayaankhan98/DSA_competative/blob/b1d200d8a9614141fa1305f1e41c605526e91336/Data_Structures/Trees/BST.cpp#L25

have a look at this code that which i had written a year ago, i hope this makes you clear what i wish to say.

```
if (root == nullptr) {
        Node* new_node = new Node;
        new_node->data = data;
        new_node->left = nullptr;
        new_node->right = nullptr;
        root = new_node;
    }
```
if the root is null then we can do something like this. @liushubin-gitHub  And i think this would be better instead of inserting the root node manually like
```
    root->val = 4;
    root->left = nullptr;
    root->right = nullptr;
```

liushubin-gitHub: Got it.Thanks a lot!

ayaankhan98: Nice look at that link i had implemented some extra methods on BST, see if you can implement that in this PR.
@liushubin-gitHub  if you face any problem let me know.

data_structures/binary_search_tree.cpp
```diff
@@ -1,172 +1,296 @@
...
 
-void Insert(node *n, int x) {
-    if (x < n->val) {
-        if (n->left == NULL) {
-            node *temp = new node;
-            temp->val = x;
-            temp->left = NULL;
-            temp->right = NULL;
-            n->left = temp;
+    if (x < root->val) {
```
ayaankhan98, line 45: what if we have a new value which is equal to the current root node? you have not handled this case when the given value is equal to the value of the current root node. we can have duplicate values in BST as far as i know there is no such boundation to not to have duplicate elements. so consider this case too.

kvedala: wouldn't that condition be treated by line# 61?

kvedala: For a test, I executed this program with this input: 
1. root = 5, insert = 5, insert = 5
2. display breadth first: `5 5 5`
3. insert = 2
4. display: `5 2 5 5`
5. insert = 7
6. display breadth first: `5 2 5 5 7`

Is this the expectation?

ayaankhan98: absolutely! thanks for pointing, my bad luck i don't know why i forgot that there is an else block

ayaankhan98: > For a test, I executed this program with the following steps:
> 
> 1. root = 5, insert = 5, insert = 5
> 2. display breadth first: `5 5 5`
> 3. insert = 2
> 4. display: `5 2 5 5`
> 5. insert = 7
> 6. display breadth first: `5 2 5 5 7`
> 7. Post order depth first: `2 7 5 5 5`
> 8. In-order depth first: `2 5 5 5 7`
> 9. Pre-order depth first: `5 2 5 5 7`
> 
> Is this the expectation?

yes these results are good

kvedala:
image
@liushubin-gitHub It seems your latest push broke the code. Please revert to the last known good code.

liushubin-gitHub:
> image
> @liushubin-gitHub It seems your latest push broke the code. Please revert to the last known good code.

OK，done。

data_structures/binary_search_tree.cpp
```diff
@@ -1,172 +1,296 @@
...
-void enqueue(node *n) { queue.t[queue.rear++] = n; }
-
-node *dequeue() { return (queue.t[queue.front++]); }
+/** insert a node to tree
+ * with greater value to right subtree, smaller value node to left subtree
+ * \param[in] root root node of a tree
+ * \param[in] x a node with value to be insert
+ */
+void Insert(node* root, int x) {
+    if (root == NULL) {
```
ayaankhan98, line 39: please consider the changes for the insert function which i suggested earlier.

liushubin-gitHub: yes.i try early.but auto format checker failed.
you can get in commit:
adb42e4

so.i revert to last good code.

ayaankhan98: @liushubin-gitHub what was the actual error? is removing the optimised code is only the solution?

liushubin-gitHub:
could non_const reference be marked a warning instead of an error
Pls review!

data_structures/binary_search_tree.cpp
ayaankhan98:
```
if (root == nullptr) {
    root = std::shared_ptr<node>(new node);
    root->val = x;
    root->left = nullptr;
    root->right = nullptr;
   return;
}

PR #933, opened 2020-07-02
Improved sieve_of_eratosthenes.cpp

iamrajiv:
#### Description of Change

Improved `sieve_of_eratosthenes.cpp` by looping till sqrt(N).

#### Checklist

Notes: Code change in `sieve_of_eratosthenes.cpp`

### DIFF
--- a/math/sieve_of_eratosthenes.cpp
+++ b/math/sieve_of_eratosthenes.cpp
@@ -25,7 +25,7 @@ bool isprime[MAX];
 void sieve(uint32_t N) {
     isprime[0] = false;
     isprime[1] = false;
-    for (uint32_t i = 2; i <= N; i++) {
+    for (uint32_t i = 2; i * i <= N; i++) {
         if (isprime[i]) {
             for (uint32_t j = (i << 1); j <= N; j += i) {
                 isprime[j] = false;
--- a/math/sieve_of_eratosthenes.cpp
+++ b/math/sieve_of_eratosthenes.cpp
@@ -15,6 +15,8 @@
 /**
  * This is the function that finds the primes and eliminates
  * the multiples.
+ * @param N number of primes to check
+ * @param [out] isprime a boolean array of size `N` identifying if `i`^th number is prime or not
  */
 void sieve(uint32_t N, bool *isprime) {
     isprime[0] = true;
--- a/math/sieve_of_eratosthenes.cpp
+++ b/math/sieve_of_eratosthenes.cpp
@@ -17,12 +17,12 @@
  * the multiples.
  */
 void sieve(uint32_t N, bool *isprime) {
-    isprime[0] = false;
-    isprime[1] = false;
+    isprime[0] = true;
+    isprime[1] = true;
     for (uint32_t i = 2; i * i <= N; i++) {
-        if (isprime[i]) {
+        if (!isprime[i]) {
             for (uint32_t j = i * i; j <= N; j = j + i) {
-                isprime[j] = false;
+                isprime[j] = true;
             }
         }
     }
@@ -33,20 +33,11 @@ void sieve(uint32_t N, bool *isprime) {
  */
 void print(uint32_t N, bool *isprime) {
     for (uint32_t i = 1; i <= N; i++) {
-        if (isprime[i]) {
+        if (!isprime[i]) {
             std::cout << i << ' ';
         }
     }
     std::cout << std::endl;
-}
-
-/**
- * Initialize the array
- */
-void init(uint32_t N, bool *isprime) {
-    for (uint32_t i = 0; i < N; i++) {
-        isprime[i] = true;
-    }
 }
 
 /**
@@ -55,7 +46,6 @@ void init(uint32_t N, bool *isprime) {
 int main() {
     uint32_t N = 100;
     bool *isprime = new bool[N];
-    init(N, isprime);
     sieve(N, isprime);
     print(N, isprime);
     delete[] isprime;
--- a/math/sieve_of_eratosthenes.cpp
+++ b/math/sieve_of_eratosthenes.cpp
@@ -21,7 +21,7 @@ void sieve(uint32_t N, bool *isprime) {
     isprime[1] = true;
     for (uint32_t i = 2; i * i <= N; i++) {
         if (!isprime[i]) {
-            for (uint32_t j = i * i; j <= N; j = j + i) {
+            for (uint32_t j = (i << 1); j <= N; j = j + i) {
                 isprime[j] = true;
             }
         }
--- a/math/sieve_of_eratosthenes.cpp
+++ b/math/sieve_of_eratosthenes.cpp
@@ -17,48 +17,48 @@
  * the multiples.
  */
 void sieve(uint32_t N, bool *isprime) {
-	isprime[0] = false;
-	isprime[1] = false;
-	for (uint32_t i = 2; i * i <= N; i++) {
-		if (isprime[i]) {
-			for (uint32_t j = i * i; j <= N; j = j + i) {
-				isprime[j] = false;
-			}
-		}
-	}
+    isprime[0] = false;
+    isprime[1] = false;
+    for (uint32_t i = 2; i * i <= N; i++) {
+        if (isprime[i]) {
+            for (uint32_t j = i * i; j <= N; j = j + i) {
+                isprime[j] = false;
+            }
+        }
+    }
 }
 
 /**
  * This function prints out the primes to STDOUT
  */
 void print(uint32_t N, bool *isprime) {
-	for (uint32_t i = 1; i <= N; i++) {
-		if (isprime[i]) {
-			std::cout << i << ' ';
-		}
-	}
-	std::cout << std::endl;
+    for (uint32_t i = 1; i <= N; i++) {
+        if (isprime[i]) {
+            std::cout << i << ' ';
+        }
+    }
+    std::cout << std::endl;
 }
 
 /**
  * Initialize the array
  */
 void init(uint32_t N, bool *isprime) {
-	for (uint32_t i = 0; i < N; i++) {
-		isprime[i] = true;
-	}
+    for (uint32_t i = 0; i < N; i++) {
+        isprime[i] = true;
+    }
 }
 
 /**
  * Main function
  */
 int main() {
-	uint32_t N = 100;
-	bool *isprime = new bool[N];
-	init(N, isprime);
-	sieve(N, isprime);
-	print(N, isprime);
-	delete[] isprime;
+    uint32_t N = 100;
+    bool *isprime = new bool[N];
+    init(N, isprime);
+    sieve(N, isprime);
+    print(N, isprime);
+    delete[] isprime;
 
-	return 0;
+    return 0;
 }
--- a/math/sieve_of_eratosthenes.cpp
+++ b/math/sieve_of_eratosthenes.cpp
@@ -32,6 +32,8 @@ void sieve(uint32_t N, bool *isprime) {
 
 /**
  * This function prints out the primes to STDOUT
+ * @param N number of primes to check
+ * @param [in] isprime a boolean array of size `N` identifying if `i`^th number is prime or not
  */
 void print(uint32_t N, const bool *isprime) {
     for (uint32_t i = 2; i <= N; i++) {
--- a/math/sieve_of_eratosthenes.cpp
+++ b/math/sieve_of_eratosthenes.cpp
@@ -32,7 +32,7 @@ void sieve(uint32_t N, bool *isprime) {
  * This function prints out the primes to STDOUT
  */
 void print(uint32_t N, bool *isprime) {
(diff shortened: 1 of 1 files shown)


kvedala:
Thank you 👍 
Please take this opportunity to improve the quality of the file with regards to:
1. documentation of the file per doxygen standards as can be seen in other files of repo [like this one](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/numerical_methods/newton_raphson_method.cpp)
2. do not delete pre-provided checks and details in the PR description.

mishraabhinn:
Is <a href="https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/math/primes_up_to_billion.cpp">this</a> is same as your PR

kvedala:
Notice that this program also has the inefficiency in memory:
There is a really large global array that is useless because of the "N" value set in the main.
The correct way to fix this would be to make the `bool *isprime` array to be a dynamic memory local to `main` function and pass it as a function argument.

kvedala:
> Is [this](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/math/primes_up_to_billion.cpp) is same as your PR

You are correct. This file is a duplicate. I suggest this file be updated with above comment and with structure and standard as in [this file](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/math/primes_up_to_billion.cpp) and the other file be deleted.

ayaankhan98:
@iamrajiv also add some test cases if you could, to verfity the code correctness.

> Notice that this program also has the inefficiency in memory:
> There is a really large global array that is useless because of the "N" value set in the main.
> The correct way to fix this would be to make the `bool *isprime` array to be a dynamic memory local to `main` function and pass it as a function argument.

and don't forget to deallocate the dynamically allocated memory.

iamrajiv:
@kvedala What you are suggesting was already implemented in PR #821 but you disapproved that and again you are suggesting the same what you disapproved can you be more precise what you want like I have already made PR #820 and #821 and closed it.

kvedala:
Yes, @iamrajiv the two pull-requests still seem the same to me. What is the difference? Please elaborate:
1. why the new pull-request and not make the requested changes to the original one
2. how is this new pull-request different from the previous one that you referenced

For future reference, please make recommended changes in the same pull-request and not create another one. This is also why we have added the details for authors to fill-up in the pull-request description that you have promptly removed and hence the conflict between the two pull-requests.

All contributors and reviewers are volunteers and hence please be considerate of their time and efforts as well.

Thank you

kvedala:
Yes, @iamrajiv the two pull-requests still seem the same to me. What is the difference?
* #820 
    * you suggested of a *different implementation*. I do not see how this is different. 
* #821 
    * suggestions recommend using a `bool` type. 
* The only *new* this pull-adds is what you have mentioned in the description - the loop termination condition.

But like another reviewer pointed out, this makes is identical to [this file](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/math/primes_up_to_billion.cpp). In which case, there is a discussion to remove the redundancy by deleting one file.

This is also why we have added the details for authors to fill-up in the pull-request description that you have promptly removed and hence the conflict between the two pull-requests.

All contributors and reviewers are volunteers and hence please be considerate of their time and efforts as well.

Thank you

kvedala:
For this pull request, may I suggest the above changes:
1. remove global array
2. allocate array dynamically to required maximum number and free after use
3. pass array to prime function

These suggested changes from comments above will make this program different from the existing ones (though not algorithmically) from an implementation perspective and then can be accepted.

@mishraabhinn and @ayaankhan98 what do you think?

iamrajiv:
@kvedala in PR #820 you have suggested that `if you would like to submit a different implementation for an existing algorithm, please submit it as a new file in the corresponding folder.` and said `You can simply close this PR and start afresh.` and because of this I have created PR #821 because previous `Sieve of Eratosthenes` code it was not using dynamic memory allocation also loop was not put sqrt(N) that's why I created a new implementation of `Sieve of Eratosthenes` in PR #821 but in that also you have disapproved and suggested to change only the for-loop to check is till sqrt(N) so that's why I created PR #933 and in that, I have changed only the for loop to check is till sqrt(N) and rest keeper the file same. 

But know you are suggesting that same implementation of PR #821 about dynamic memory allocation which was disapproved by you.

Anyways, I will implement what you have suggested and also make sure the code is well-formatted according to #807 and [this file](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/math/primes_up_to_billion.cpp).

iamrajiv:
@kvedala should I remove `init` function because bool array is by default initialise with all N element with `false` so no need to again initialize with `true` rather we can change `sieve` function according to it something like this
```
void sieve(uint32_t N, bool *isprime) {
    isprime[0] = true;
    isprime[1] = true;
    for (uint32_t i = 2; i * i <= N; i++) {
        if (!isprime[i]) {
            for (uint32_t j = i * i; j <= N; j = j + i) {
                isprime[j] = true;
            }
        }
    }
}
```

mishraabhinn:
> @kvedala should I remove `init` function because bool array is by default initialise with all N element with `false` so no need to again initialize with `true` rather we can change `sieve` function according to it something like this
> 
> ```
> void sieve(uint32_t N, bool *isprime) {
>     isprime[0] = true;
>     isprime[1] = true;
>     for (uint32_t i = 2; i * i <= N; i++) {
>         if (!isprime[i]) {
>             for (uint32_t j = i * i; j <= N; j = j + i) {
>                 isprime[j] = true;
>             }
>         }
>     }
> }
> ```

Agreed.

mishraabhinn:
> @kvedala should I remove `init` function because bool array is by default initialise with all N element with `false` so no need to again initialize with `true` rather we can change `sieve` function according to it something like this
> 
> ```
> void sieve(uint32_t N, bool *isprime) {
>     isprime[0] = true;
>     isprime[1] = true;
>     for (uint32_t i = 2; i * i <= N; i++) {
>         if (!isprime[i]) {
>             for (uint32_t j = i * i; j <= N; j = j + i) {
>                 isprime[j] = true;
>             }
>         }
>     }
> }
> ```

It will create ambiguity.

ayaankhan98:
@mishraabhinn can you please explain the ambiguity that you mentioned?

math/sieve_of_eratosthenes.cpp
```diff
@@ -49,17 +43,22 @@ void print(uint32_t N) {
 /**
  * Initialize the array
  */
-void init() {
-    for (uint32_t i = 1; i < MAX; i++) {
+void init(uint32_t N, bool *isprime) {
```
kvedala, line 46: Yes, this function cam be removed

math/sieve_of_eratosthenes.cpp
```diff
@@ -49,17 +43,22 @@ void print(uint32_t N) {
...
+/**
+ * Main function
+ */
 int main() {
     uint32_t N = 100;
-    init();
-    sieve(N);
-    print(N);
+    bool *isprime = new bool[N];
+    init(N, isprime);
```
kvedala, line 58:
```suggestion
    memset(isprime, true, N * sizeof(bool)); // preset all numbers to `true`
```
this on-line is same as what the `init` function does above. Hence, if you use this line, you can delete the init function.
You might have to include `<cstring>` to use this function.

math/sieve_of_eratosthenes.cpp
```diff
@@ -10,24 +10,18 @@
...
  */
-void sieve(uint32_t N) {
+void sieve(uint32_t N, bool *isprime) {
     isprime[0] = false;
     isprime[1] = false;
-    for (uint32_t i = 2; i <= N; i++) {
+    for (uint32_t i = 2; i * i <= N; i++) {
         if (isprime[i]) {
-            for (uint32_t j = (i << 1); j <= N; j += i) {
+            for (uint32_t j = i * i; j <= N; j = j + i) {
```
kvedala, line 24: The original implementation was correct. 
* `i<<1` -> this means, start from `i*2` and mark all multiples of i as not-prime
* What you have done: `i*i` -> this means, start from `i*i` and mark all multiples of i as not-prime. Thus, you have lost all numbers from `i*2` to `i*i`!

kvedala:
> @mishraabhinn can you please explain the ambiguity that you mentioned?

did you mean to imply that:
* `init` function initialized all memory to `true` = 1 while
* the default value set is 0 = `false`? 
Hence, a memory is is required. But can be done so more efficiently as mentioned above.

mishraabhinn:
> > @mishraabhinn can you please explain the ambiguity that you mentioned?
> 
> did you mean to imply that:
> 
> * `init` function initialized all memory to `true` = 1 while
> * the default value set is 0 = `false`?
>   Hence, a memory is is required. But can be done so more efficiently as mentioned above.

ok.

math/sieve_of_eratosthenes.cpp
iamrajiv: @kvedala I am not getting the difference between using `i << 1` and `i * i` since I am getting same output for both the cases when I executed the program locally.

iamrajiv: @kvedala I am not getting the difference between using `i << 1` and `i * i` since I am getting the same output for both the cases when I executed the program locally.

kvedala: the difference is as mentioned above:
1. the inner loop start is `i^2` (i squared) in your case and `i*2` (i times 2) in the original implementation
2. In the original implementation, the outer loop will skip for many numbers less than `i*i` because the inner loop will mark the non-primes. Hence, the inner loop will not execute for all numbers
3. In your case, he outer loop will check every number less then i squared and hence, the inner loop will execute for every number!

If it is still not clear, try to work the loop for i 23 on a piece of paper using the above algorithm step by step or use cout statements and check every possibility.

iamrajiv: Oh I got it you mean for example I = 2 it will skip 2 and start from 4 according to my code is it right @kvedala

iamrajiv: Oh, I got it you mean for example i = 2 it will skip 2 and start from 4 according to my code is it right @kvedala

kvedala: yes, @iamrajiv

math/sieve_of_eratosthenes.cpp
```diff
@@ -37,29 +31,24 @@ void sieve(uint32_t N) {
 /**
  * This function prints out the primes to STDOUT
  */
-void print(uint32_t N) {
+void print(uint32_t N, bool *isprime) {
     for (uint32_t i = 1; i <= N; i++) {
```
kvedala, line 35: 1 is neither prime nor composite.
I understand the original code has the erro. But since the code is being fixed, lets fix it for good and fix all problems that exist :)
`i` should start from `2`

math/sieve_of_eratosthenes.cpp
kvedala: make this fix as well.

math/sieve_of_eratosthenes.cpp
```diff
@@ -37,29 +31,24 @@ void sieve(uint32_t N) {
 /**
  * This function prints out the primes to STDOUT
  */
-void print(uint32_t N) {
-    for (uint32_t i = 1; i <= N; i++) {
-        if (isprime[i]) {
+void print(uint32_t N, bool *isprime) {
```
kvedala, line 34:
```suggestion
void print(uint32_t N, const bool *isprime) {
```

kvedala:
See [this comment](https://github.com/TheAlgorithms/C-Plus-Plus/pull/933#discussion_r452532445)

math/sieve_of_eratosthenes.cpp
iamrajiv: @kvedala I am not getting it should I revert the code to `for (uint32_t j = (i << 1); j <= N; j += i)`

kvedala: yes

math/sieve_of_eratosthenes.cpp
```diff
@@ -10,25 +10,19 @@
...
-#define MAX 10000000
-
-/** array to store the primes */
-bool isprime[MAX];
+#include <iostream> // for io operations
 
 /**
  * This is the function that finds the primes and eliminates
  * the multiples.
  */
```
kvedala, line 18:
```suggestion
 * @param N number of primes to check
 * @param [out] isprime a boolean array of size `N` identifying if `i`^th number is prime or not
 */
```

math/sieve_of_eratosthenes.cpp
```diff
@@ -37,29 +31,24 @@ void sieve(uint32_t N) {
 /**
  * This function prints out the primes to STDOUT
  */
```
kvedala, line 33:
```suggestion
 * @param N number of primes to check
 * @param [in] isprime a boolean array of size `N` identifying if `i`^th number is prime or not
 */
```

PR #954, opened 2020-07-16
Fixgraph : improvement in remaning graph Algorithms

ayaankhan98:
#### Description of Change
improvement in remaning graph Algorithms

#### Checklist

- [x] Added description of change
- [x] Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#New-File-Name-guidelines)
- [x] Added tests and example, test must pass
- [x] Added documentation so that the program is self-explanatory and educational - [Doxygen guidelines](https://www.doxygen.nl/manual/docblocks.html)
- [x] Relevant documentation/comments is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#Commit-Guidelines)
- [x] Search previous suggestions before making a new one, as yours may be a duplicate.
- [x] I acknowledge that all my contributions will be made under the project's license.

Notes: 

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/954"></a>

PR #932, opened 2020-07-01
fix: graph algorithm improvements

ayaankhan98:
#### Description of Change

issue related to this #909
- Code quality Improvement over graph Algorithms
- Added Documentation
- Added CMakeLists.txt

#### Checklist

- [x] Added description of change
- [x] Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#New-File-Name-guidelines)
- [x] Added tests and example, test must pass
- [x] Added documentation so that the program is self-explanatory and educational - [Doxygen guidelines](https://www.doxygen.nl/manual/docblocks.html)
- [x] Relevant documentation/comments is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#Commit-Guidelines)
- [x] Search previous suggestions before making a new one, as yours may be a duplicate.
- [x] I acknowledge that all my contributions will be made under the project's license.

Notes:

PR #981, opened 2020-07-27
Refactor lowest common ancestor

fhlasek:
#### Description of Change

Making the code more modular, following the coding guidelines, flexible, documented and tested.

#909 
#980 

#### Checklist

- [X] Added description of change
- [X] Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#New-File-Name-guidelines)
- [X] Added tests and example, test must pass
- [X] Added documentation so that the program is self-explanatory and educational - [Doxygen guidelines](https://www.doxygen.nl/manual/docblocks.html)
- [X] Relevant documentation/comments is changed or added
- [X] Search previous suggestions before making a new one, as yours may be a duplicate.
- [X] I acknowledge that all my contributions will be made under the project's license.

Notes: 

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/981"></a>

PR #916, opened 2020-06-26
[fix/docs]: Update backtracking folder

Panquesito7:
#### Description of Change

* Add `backtracking` folder to CMake build-chain
* Update various files in the `backtracking` folder with refined code and documentation

#### Checklist

- [x] Added description of change
- [x] Relevant documentation/comments is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#Commit-Guidelines)
- [x] Search previous suggestions before making a new one, as yours may be a duplicate.
- [x] I acknowledge that all my contributions will be made under the project's license.

Notes: Didn't split commits. If requested, I can split, but will take more time. 

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/916"></a>

PR #986, opened 2020-08-03
fhlasek/fixgraph: refactor of graph/ to follow the linter style

fhlasek:
#### Description of Change
Continuation of #954 and #983. Addresses #932.

#### Checklist

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/986"></a>


graph/bfs.cpp
```diff
@@ -1,62 +1,196 @@
...
+ */
+namespace graph {
+/**
+ * \brief
+ * Adds and edge between two vertices of graph say u and v in this
+ * case.
+ *
+ * @param adj Adjacency list representation of graph
+ * @param u first vertex
+ * @param v second vertex
```
Panquesito7, line 64: Missing `@return` statement (documentation).

graph/bfs.cpp
```diff
@@ -1,62 +1,196 @@
...
-            if (visited[*it] == false) {
-                visited[*it] = true;
-                q.push_back(*it);
+
+/**
+ * \brief
+ * Function performs the breadth first search algorithm over the graph
+ *
+ * @param adj Adjacency list representation of graph
+ * @param start vertex from where traversing starts
```
Panquesito7, line 84: Missing `@return` statement (documentation).

Panquesito7:
Please, pick one file at a time, not more than five.
We will do much better job improving file by file.

Thank you.

kvedala:
> Please, pick one file at a time, not more than five.
> We will do much better job improving file by file.
> 
> Thank you.

Practice before you preach. So far, contributions by @fhlasek have been excellent with minimal corrections required. Your contribution #916 has surpassed over 100 comments and suggestions!

graph/bfs.cpp
kvedala: These functions return `void`. At minimum, one of `@param` or `@return` is required. In the two cases pointed by you, that condition is satisfied.

kvedala:
@fhlasek I think this is a work in progress while the CI errors are resolved. Can you please mark the pull-request as `draft` until it is ready for review to avoid reviews of intermediate code. Thanks.

fhlasek:
@Panquesito7 thanks for the suggestions. I definitely prefer to make smaller incremental changes. However, in this case, when I attempted it (#981) it was suggested to me to merge the work with this mega PR (#954).

I guess there's not an easy and clear solution since this part of the codebase is not compiled and run as part of the CI - that makes it hard to make any modifications and small incremental improvements in individual files. On the other hand, adding the CMakefile to compile the folder introduces a lot of CI failures as the code does not follow the coding guidelines and before merging, a lot of code needs to be modified to fix the CI errors.

I'm happy to continue either way, whichever the maintainers consider preferable, and even more happy to learn some new approach to make these sort of refactoring easy to review :-) Looking to get some guidance as I'm fairly new to the opensource community!

@kvedala Thanks for the encouraging words! That's exactly the reason I'm trying to help. CI is technically passing, but there are still some warnings from the LGTM about too short variable names. I'll see if I can resolve these easily, then take a final sweep and will try to get it ready to merge as soon as possible so that we can then continue on further improvements in smaller incremental PRs :-)

fhlasek:
@kvedala, @Panquesito7 this PR does not fix all the issues (not even aiming to) - for example, code is lacking documentation, defines global static arrays, etc. However, it passes the CI which should enable other developers to contribute. I propose to merge it as it is and continue improving the code quality in small incremental changes. What do you think?

kvedala:
@fhlasek Sounds good. A couple of files at a time might be better. @ayaankhan98 also suggested the same and seems to be more productive this way.
However, please note that existing documentations and fixes got removed in this PR. For example, check `cycle_check_directed_graph`.

fhlasek:
@kvedala thanks for the suggestion, I'm close to giving up, but will give it one more final go :-) I'm especially frustrated that I was asked to do something and after finishing the work, I'm being told it's not the right way and that I should do it differently. Finally, most of the code is not even mine as I only carried the work of @ayaankhan98 (#954). As I said, I'm happy to own the work and carry it forward, but I'd also appreciate a bit of compassion ;-)

ayaankhan98:
> @kvedala thanks for the suggestion, I'm close to giving up, but will give it one more final go :-) I'm especially frustrated that I was asked to do something and after finishing the work, I'm being told it's not the right way and that I should do it differently. Finally, most of the code is not even mine as I only carried the work of @ayaankhan98 (#954). As I said, I'm happy to own the work and carry it forward, but I'd also appreciate a bit of compassion ;-)

yes you are right modifying the code written by other is a bit frustrating, most of the code linked in this PR is not written by me, i was also on the task of improving these graph algorithms code. i worked on some files but eventually i gave up.

kvedala:
> @kvedala thanks for the suggestion, I'm close to giving up, but will give it one more final go :-) I'm especially frustrated that I was asked to do something and after finishing the work, I'm being told it's not the right way and that I should do it differently. Finally, most of the code is not even mine as I only carried the work of @ayaankhan98 (#954). As I said, I'm happy to own the work and carry it forward, but I'd also appreciate a bit of compassion ;-)

I sympathize with you @fhlasek. We have come a long way from having no code-coverage, co compilation, no documentation to a much more refined code repository and thanks to a lot of contributors and @ayaankhan98 was instrumental in making this happen. 

Let us not give up, especially when we are so close to making all all code follow the protocols 💯  and it is not and has not been a one man or a one pull-request or a one-day job. We will have to achieve this collectively and it has been going great, keep it up 👍  😄

kvedala:
@fhlasek, 
Please note that @ayaankhan98  and myself will review this PR as you both have already done much of the work. No need to go back & forth.@ayaankhan98 can you confirm. Thanks.

ayaankhan98:
> @fhlasek,
> Please note that @ayaankhan98 and myself will review this PR as you both have already done much of the work. No need to go back & forth.@ayaankhan98 can you confirm.
> 
> The other Pull-requests though closed, can remain so. We will review only the files that have been marked for update. The other un-touched files can be updated in a subsequent pull-request.
> 
> Please let know if this sounds like a decent plan.
> 
> Thanks.

sounds good! :+1:

fhlasek:
Sounds awesome, thanks @kvedala and @ayaankhan98 !

Let the perfection not be the enemy of the good. The current state is pretty bad and if we can quickly take it to the state where we can run the CI, I believe it would be a big win even if we caused some minor regressions or imperfections in return. Honestly, I almost think that at this stage a thorough code review might not even be worth it.

In the meanwhile, I started splitting of some of the files #990, #991, #992, #993, #994, there might be easier to review and resolve, but there's not really any CI running for them.

kvedala:
I am liking this PR than the individual ones A lot of changes and all passing the CI 🙂 
Is it ok if I review this over the weekend? @ayaankhan98 feel free to review meanwhile.

graph/dfs.cpp
```diff
@@ -1,26 +1,133 @@
+/**
```
kvedala, line 1: This file can be renamed to maybe `depth_first_search.cpp` or `depth_first_search_dfs.cpp`?

kvedala:
There is another problem here:
There are many repeated functions and all in the same namespace. For example: `addEdge`. Might be asking too much but would it be better to have a `.h` file with most of the common functions here and not repeat them in individual files? This will also help maintain the reference links in the generated documentation.

graph/dfs.cpp
fhlasek: good point, I'll rename it

fhlasek:
@kvedala I was thinking about it - specifically, it might be beneficial to have a common data structure to store a graph and all other algorithms using it as a way to take input.

However, then I noticed a guideline in the intro of this repo:

> Each source code is atomic using STL classes and no external libraries are required for their compilation and execution. Thus the fundamentals of the algorithms can be studied in much depth.

What do you think? I believes the rules are meant to be broken if they are making the code unnecessarily complicated / repetitive.

kvedala:
We are neither breaking nor bending the rules. 
By keeping the headerfile, we are still keeping the entire code atomic. When I wrote that line in the README, I mean that nowehre in the repo we are relying on third party libraries that need to be linked. For example, there are many (and quite standard) libraries for matrix operations but all code here rely on code from the repo and mostly from the same folder. If there are any dependencies, they are only by one file atmost. 

For example, when I wrote the QR and LU algorithms and eigen value computations that use funadamental algorithms in a header file. But for inverse computation in Hill cipher, I only call one function from the header file while still rewriting all code for matrix inverse again, even though it already exists in another program. The header files are serving only specific purpose only as the description and name suggest. They too are atomic in implementation and hence any person reviewing the code would be able to conveniently navigate through the code.

fhlasek:
@kvedala cool, makes sense. I think we should do it then. However, I believe that might be a rather large refactor that I'd prefer to keep separate from this effort to enable CI for this subfolder. What do you think?

kvedala:
> @kvedala cool, makes sense. I think we should do it then. However, I believe that might be a rather large refactor that I'd prefer to keep separate from this effort to enable CI for this subfolder. What do you think?

This PR is currently showing 16 files modified; while there are also parts of this PR being merged as individual PRs. Hence, towards the end, the merge conflicts with this PR will need to be carefully resolved. When so, I expect the number of modified files would be significantly less. 

That being said, the addition of a header file like `graph_class.h` would be a much smaller change. Now that I am thinking out loudly, what you suggest does make sense. It would be better to make that change in a new PR and leave the code as is for now. @ayaankhan98 thoughts?

ayaankhan98:
@kvedala  and @fhlasek yes, it sounds good to me too, to do the code refactoring in the subsequent PR's and let this PR for a sole purpose of enabling CI for graph

Panquesito7:
Sounds good to me. We can merge this to enable CI and continue
improving the code and adding documentation in another PR. 👍

fhlasek:
@Panquesito7 @kvedala @ayaankhan98 do you want me to continue splitting files one by one or do you think it's manageable at the current size?

kvedala:
> @Panquesito7 @kvedala @ayaankhan98 do you want me to continue splitting files one by one or do you think it's manageable at the current size?

Maybe we can split a couple of more files, so that this repo contains a single digit of modified files? Thank you @fhlasek

graph/depth_first_search.cpp
```diff
@@ -0,0 +1,133 @@
...
+ * Adds and edge between two vertices of graph say u and v in this
+ * case.
+ *
+ * @param adj Adjacency list representation of graph
+ * @param u first vertex
+ * @param v second vertex
+ *
+ */
+void addEdge(std::vector<std::vector<size_t>> *adj, size_t u, size_t v) {
+    /**
```
kvedala, line 57:
```suggestion
    /*
```

graph/cycle_check_directed_graph.cpp
```diff
@@ -65,8 +65,8 @@ class Graph {
      * @param vertices specify the number of vertices the graph would contain.
      * @param adjList is the adjacency list representation of graph.
      */
-    Graph(unsigned int vertices, AdjList const& adjList)
-        : m_vertices(vertices), m_adjList(adjList) {}
+    Graph(unsigned int vertices, AdjList adjList)
```
kvedala, line 68: this change I think reverted a good code. please check

kvedala: this change I think reverted a good code. please check

kvedala, line 68: did this line revert a good code?

graph/depth_first_search.cpp
kvedala, line 57:
```suggestion
    /*
```

graph/cycle_check_directed_graph.cpp
fhlasek: This is actually what the linter suggests:
```
/Users/filip/github/C-Plus-Plus/graph/cycle_check_directed_graph.cpp:68:34: warning: pass by value and use std::move [modernize-pass-by-value]
    Graph(unsigned int vertices, AdjList const& adjList)
                                 ^~~~~~~~~~~~~~
                                 AdjList
```

fhlasek: if I understand correctly, the input parameter is copied regardless, and therefore it's encouraged to copy it directly in the parameter passing since the compiler can better optimise

kvedala: Got it. [Reference to more on this](https://stackoverflow.com/questions/26261007/why-is-value-taking-setter-member-functions-not-recommended-in-herb-sutters-cpp)

graph/depth_first_search.cpp
fhlasek: not sure what you want me to change here. Multi-line comments starting with `/**` seems to be a very common style all across the codebase. (https://www.doxygen.nl/manual/docblocks.html)

kvedala: yes, but not inside the function. If done so, the comment within the block gets displayed and will not render properly. When commenting inside a function, use `/*` instead of using `/**`.

fhlasek: Oh, good catch. This is not my code and didn't notice that. Will fix it ;-)

ayaankhan98 (approved):

ayaankhan98:
great work @fhlasek 
Thanks for your contribution!

PR #990, opened 2020-08-05
fix: refactor BFS algorithm.

fhlasek:
#### Description of Change
Splitting up from the original commit from #954 (@ayaankhan98)
Note that there's no CI for this folder yet.

#### Checklist

- [X] Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#New-File-Name-guidelines)

Notes: 

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/990"></a>


Panquesito7:
Good work updating the code. 👍

> Note that there's no CI for this folder yet.

Please follow the next steps:

1. Copy `CMakeLists.txt` from the `math` folder (keep the file name as-is).
2. Paste the copied file in the `graph` folder.
2. On line 16, change from `bin/math` to `bin/graph`.
3. Open `CMakeLists.txt` in the root directory. 
4. Next, add the following code between `geometry` and `graphics`:

```CMake
add_subdirectory(graph)
```

That's it. Commit and push your changes.
Next, we will confirm that the compilation works properly. 🙂

You can use the [Gitpod](https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/990) badge above and compile the code online.

Panquesito7:
> - [ ] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#Commit-Guidelines)

Please follow commit guidelines.
It is specified in the [contributing guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#Commit-Guidelines).

graph/breadth_first_search.cpp
```diff
@@ -0,0 +1,185 @@
...
+ * of the graph.
+ *
+ * @param graph Adjacency list representation of graph
+ * @param u first vertex
+ * @param v second vertex
+ *
+ */
+void add_undirected_edge(std::vector<std::vector<int>> *graph, int u, int v) {
+    add_directed_edge(graph, u, v);
+    add_directed_edge(graph, v, u);
```
Panquesito7, line 84: You must add `graph::` at the start of functions that are in the `graph` namespace.
Even if you are registering another function inside the same namespace, it won't work without it.
```suggestion
    graph::add_directed_edge(graph, u, v);
    graph::add_directed_edge(graph, v, u);
```

Panquesito7:
When renaming files, please use the `git move` operation as this will keep file history.
If you remove the file, contributor's and file history will be lost.

fhlasek:
@Panquesito7 thanks for super fast review!

> Note that there's no CI for this folder yet.

That's the entire purpose of doing this refactor. If we added the `CMakeList` CI would start immediately failing because of so many issues in the entire directory. We need to either fix them all at once in a single large PR (#986) or this way without running the CI.

> In the future, please follow commit guidelines.
It is specified in the contributing guidelines.

Thanks, will try to do a better job.

> When renaming files, please use the git move operation as this will keep file history.
If you remove the file, contributor's and file history will be lost.

I actually used `git move` to move this file. I believe it's only the github view that's for convenience displaying the changes as removed and added a new file since there is also a lot of changes. You can actually see that the contributions history of the file is maintainer correctly: https://github.com/TheAlgorithms/C-Plus-Plus/blame/5450be08cf607938620e351dbc32f07c4949e819/graph/breadth_first_search.cpp

graph/breadth_first_search.cpp
fhlasek: That is factually incorrect. This is within the language specification, and is used in this very codebase without violating any linter rules (e.g. https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/data_structures/skip_list.cpp#L26)

Panquesito7:
> fix them all at once

If we do so, we shall edit one file at a time.
So, we update them as best as possible.

> used `git move` to move this file

Good work, thank you for doing this.
I've just reviewed it, looks good. 👍

graph/breadth_first_search.cpp
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * \brief
+ * Function performs the breadth first search algorithm over the graph
+ *
+ * @param graph Adjacency list representation of graph
+ * @param start vertex from where traversing starts
+ * @returns a binary vector indicating which vertices were visited during the search.
+ *
+ */
+std::vector<bool> beadth_first_search(const std::vector<std::vector<int>> &graph,
```
Panquesito7, line 96: Is this the correct function name?
Isn't it `breadth_first_search`?

The file's name is the same as above.

graph/breadth_first_search.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include <iostream>
+#include <queue>
+#include <vector>
+
+/**
+ * \namespace graph
+ * \brief Graph algorithms
+ */
+namespace graph {
+
```
Panquesito7, line 57: Please remove this empty line.

fhlasek: will do! :-)

graph/breadth_first_search.cpp
fhlasek: good spot, that's definitely a typo

Panquesito7:
@kvedala please review as well. Thanks.

kvedala:
I like the code as well. My only concern is the following:
> the existing code, though not conforming to the clang-tidy specs in place, provides an alternative implementation. By merging this PR, we will loose that implementation. Whenever we find two implementations for the same algorithm, we are keeping both the implementations so that the readers can learn from both. so, `breadth_search_first1.cpp` and `breadth_search_first2.cpp` would have been the way to go. 

@fhlasek  Is there any chance to salvage the existing implementation as well? Thanks.

fhlasek:
@kvedala that's a good point, I definitely support that especially in cases where there are two possible ways to implement the algorithm. In this specific case the implementation is pretty much identical with the difference that the original one is quite poorly written and does not follow any good coding practices and guidelines. In this case, if we tried to keep both implementations in a decent shape, we'd likely end up with almost identical code.

PR #1023, opened 2020-08-14
test: add test for fibonacci() function

fzxutar:
#### Description of Change
- Added a test function for the fibonacci() function.

#### Checklist

Notes: 
A simple test function was added to test the fibonacci() function to make sure that the results are correct.

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1023"></a>

### DIFF
--- a/math/fibonacci.cpp
+++ b/math/fibonacci.cpp
@@ -24,8 +24,39 @@ int fibonacci(unsigned int n) {
     return fibonacci(n - 1) + fibonacci(n - 2);
 }
 
+/**
+ * Function for testing the fibonacci() function with a few
+ * test cases and assert statement.
+ */
+void test() {
+    int test_case_1 = fibonacci(0);
+    assert(test_case_1 == 0);
+    std::cout << "Passed Test 1!" << std::endl;
+
+    int test_case_2 = fibonacci(1);
+    assert(test_case_2 == 1);
+    std::cout << "Passed Test 2!" << std::endl;
+
+    int test_case_3 = fibonacci(2);
+    assert(test_case_3 == 1);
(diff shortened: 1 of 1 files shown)


kvedala:
It seems that the error is from the existing code and not in your contribution. 
As the error states, the function `fibonacci()` is returning an `unsigned int` while its definition for return type is `int`. 
This can be rectified by changing the function return type on line 17:
```cpp
unsigned int fibonacci(unsigned int n) {
```
image

math/fibonacci.cpp
```diff
@@ -24,8 +24,39 @@ int fibonacci(unsigned int n) {
     return fibonacci(n - 1) + fibonacci(n - 2);
 }
 
+/**
+ * Function for testing the fibonacci() function with a few
+ * test cases and assert statement.
+ */
+void test() {
```
kvedala, line 31:
```suggestion
static void test() {
```
This is to specify that the function is only for this particular file and not to be globally visible.

math/fibonacci.cpp
```diff
@@ -24,8 +24,39 @@ int fibonacci(unsigned int n) {
     return fibonacci(n - 1) + fibonacci(n - 2);
 }
 
+/**
+ * Function for testing the fibonacci() function with a few
+ * test cases and assert statement.
+ */
```
kvedala, line 30:
```suggestion
  * @returns `void`
  */
```
This is needed for the documentation generator to recognize `test()` as a function.

fzxutar: Thanks for the suggestions. I fixed the errors and now all checks have passed.

Issue #1004, opened 2020-08-12
[FEATURE] Build a common graph representation


fhlasek:
Most of the algorithms in the graph folder implement their own representation of the graph structure and very often it is quite similar piece of the code.

## Detailed Description

- [ ] Build a new header-only library in `graph` folder that will contain a class for storing a graph as an adjacency list.
- [ ] Use the structure to pass the graph object as a parameter for graph algorithms in the folder

## Context
Pointed out in #986 as a desirable thing to have to reduce the code duplication and also contributes to #909 

## Possible Implementation
I believe the implementation in [cycle_check_directed_graph.cpp](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/graph/cycle_check_directed_graph.cpp#L54) might be a good start

PR #1065, opened 2020-08-29
feat: graph data structure.

fhlasek:
#### Description of Change
#1004 The data structure can be further enhanced by adding weighted edges or more functionality for various constructors or IO operations. Aiming for the very minimal usable skeleton to make it easier to review.

#### Checklist

- [X] Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#New-File-Name-guidelines)
- [X] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#Commit-Guidelines)

Notes: 

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1065"></a>


fhlasek:
@kvedala can you please help me here? It seems like the code checker is aiming to process the file with C language rather than C++
```
error: invalid argument '-std=c++11' not allowed with 'C' [clang-diagnostic-error]
/home/runner/work/C-Plus-Plus/C-Plus-Plus/data_structures/graph.h:21:10: error: 'cassert' file not found [clang-diagnostic-error]
#include <cassert>
         ^
```

kvedala:
1. `.h` file is considered by `clang` as a C-file and not a CPP. Changing the `graph.h` to `graph.hpp` might work. 
2. the filenames are ambiguous and might be better to use `directed_graph.h` and `directed_graph.cpp`. But see point 3 below
3. `graph.h` is simple enough to be included in the cpp file itself. Hence, a single `directed_graph.cpp might suffice here?

fhlasek:
@kvedala thanks for the suggestion, renaming to `.hpp` seems to work. I don't quite follow what the other recommendations are. Are you happy with how the code is organised right now or do you want me to change it somehow?

kvedala:
> @kvedala thanks for the suggestion, renaming to `.hpp` seems to work. I don't quite follow what the other recommendations are. Are you happy with how the code is organised right now or do you want me to change it somehow?

I that a previous discussion led to us thinking that having a `graph.h` file generic to all might be a good way to go. However, thinking more about it, it might actually cause more issues to contributors. It is really difficult to maintain a generic `graph.h` code that would cater every contributor's needs and for the diverse graph algorithms there are but would not be able to readily use the header file.

fhlasek:
@kvedala I suggest we check the code in, but don't enforce using this structure for every graph algorithm. I understand that certain algorithms might not be an easy fit for the simple structure representation, and they'll have a full discretion to define their own input structure. I believe this way we can greatly reduce the code duplication and make it actually much easier to maintain while keeping the flexibility. What do you think?

fhlasek:
Alternatively, since the main purpose of this repo is educational, we can have a very minimalistic version of this common piece of code in form of a _oneliner_ typedef that'd define the graph structure as an adjacency list.

fhlasek:
@kvedala do you think #1076 would be a better approach?

kvedala:
Sure @fhlasek 
1. Including the in the repo does no harm and we can proceed with the merge.
2. Having a skeleton file as in this repo is fine but need not be enforced to be used by the contributors. 
3. The approach in #1076 is good and in accordance with modern C++ practices. Again, need not be enforced.

data_structures/test_graph.cpp
```diff
@@ -0,0 +1,53 @@
+#include <cassert>
+#include <set>
+
+#include "./graph.hpp"
+
+/** Test function */
```
kvedala, line 6:
```suggestion
/** Test function 
 * @returns void
 */
```

data_structures/graph.hpp
```diff
@@ -0,0 +1,119 @@
...
+ */
+namespace data_structures {
+
+/**
+ * @brief Data structure storing a directed graph as an adjacency list.
+ */
+class directed_graph {
+ public:
+    /** @brief For every vertex a list of its neigbors is stored.  */
+    std::vector<std::list<int> > neighbors;
```
kvedala, line 37: should this be made private?

PR #1075, opened 2020-09-05
feat: Added Freivald's algorithm

Darshana-Sarma:
#### Checklist

- [ ] Added description of change

Notes: 

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1075"></a>

### DIFF
--- a/others/Coppersmith_frievald's.cpp
+++ b/others/Coppersmith_frievald's.cpp
@@ -1,0 +1,154 @@
+#include<iostream>
+#include<stdio.h>
+#include<stdlib.h>
+
+using namespace std;
+
+int main()
+{
+    int n;
+    cout<<"Enter the dimension of the matrices: ";
+    cin>>n;
+    double a[n][n];
+    cout<<"\nEnter the 1st matrix: \n";
+    for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<n;j++)
+        {
+            cout<<"a["<<i<<"]["<<j<<"]= ";
+            cin>>a[i][j];
+        }
+    }
+     double b[n][n];
+    cout<<"\nEnter the 2nd matrix: \n";
+    for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<n;j++)
+        {
+            cout<<"b["<<i<<"]["<<j<<"]= ";
+            cin>>b[i][j];
+        }
+    }
+    cout<<"\nThe first matrix is:\n ";
+    for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<n;j++)
+            cout<<a[i][j]<<"\t";
+            cout<<"\n";
+    }
+    cout<<"\n";
+    cout<<"\nThe Second matrix is:\n ";
+    for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<n;j++)
+            cout<<b[i][j]<<"\t";
+            cout<<"\n";
+    }
+    cout<<"\n";
+
+    double c[n][n];
+    cout<<"Enter the result matrix: \n";
+    for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<n;j++)
+        {
+            cout<<"c["<<i<<"]["<<j<<"]= ";
+            cin>>c[i][j];
+        }
+    }
+    cout<<"The resultant matrix: \n";
+    for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<n;j++)
+        {
+            cout<<c[i][j]<<"\t";
+        }
+        cout<<endl;
+    }
+    cout<<"\nrandom generation of the 'r' vector containing only 0/1 as it's elements\n";
+    double r[n][1];
+    for(int i=0;i<n;i++)
+    {
+        r[i][0]=rand() % 2;
+        cout<< r[i][0]<< "\n";
+    }
+    cout<<"\nTest: a * (b*r) - (c*r) = 0\n";
+    double br[n][1];
+    for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<1;j++)
+        {
+            for(int k=0;k<n;k++)
+            br[i][j]=br[i][j]+ b[i][k]*r[k][j];
+
+        }
+    }
+    double cr[n][1];
+    for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<1;j++)
+        {
+            for(int k=0;k<n;k++)
+                cr[i][j]=cr[i][j]+c[i][k]* r[k][j];
+        }
+    }
+    double abr[n][1];
+    for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<1;j++)
+        {
+            for(int k=0;k<n;k++)
+                abr[i][j]=abr[i][j]+a[i][k]*br[k][j];
+        }
+    }
+    /*cout<<"\nMatrix b*r \n";
+     for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<1;j++)
+        {
+            for(int k=0;k<n;k++)
+            cout<<br[i][j]<<"\n";
+        }
+    }
+    cout<<"\nMatrix c*r \n";
+    for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<1;j++)
+        {
+            for(int k=0;k<n;k++)
+                cout<<cr[i][j]<<"\n";
+        }
+    }
+    cout<<"\nMatrix a*(b*r) \n";
+    for(int i=0;i<n;i++)
+    {
+        for(int j=0;j<1;j++)
+        {
+            for(int k=0;k<n;k++)
+                cout<<abr[i][j]<<"\n";
+        }
+    }*/
+    //abr-cr
+    for(int i=0;i<n;i++)
+    {
+        abr[i][0]=abr[i][0]-cr[i][0];
+    }
+    cout<<"\nResult of a*(b*r)-(c*r): \n";
+     for(int i=0;i<n;i++)
+    {
+        cout<<abr[i][0]<<"\n";
+    }
+
+    bool flag = true;
+    for(int i=0;i<n;i++)
+    {
+        if(abr[i][0]==0)
+            continue;
+        else
+            flag = false;
+    }
+    if (flag == true)
+        cout<< "\nYes,'c' is the product of matrix a and b  ";
+    else
+        cout<< "\nNo";
+}
--- a/others/Coppersmith_frievald's.cpp
+++ b/others/Coppersmith_frievald's.cpp
@@ -1,154 +1,163 @@
-#include<iostream>
-#include<stdio.h>
-#include<stdlib.h>
+/**
+ * @file
+ * @brief Implementation of [Freivalds' algorithm]
+ * (https://en.wikipedia.org/wiki/Freivalds%27_algorithm)
+ * @author [Darshana Sarma](https://github.com/Darshana-Sarma)
+ * @details
+ * Given three n Ã— n matrices A, B, and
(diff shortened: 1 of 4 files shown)


others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,163 @@
...
+ *runtime of O(kn^{2})and error probability of <= 1/(2^k) is achieve
+ */
+
+/**
+ * Test function 1 with A = [[1,0],[2,1]]
+ * B = [[5,4],[-5,1]]
+ * B = [[5,4],[5,9]]
+ * @returns none
+ */
+
```
Panquesito7, line 122:
```suggestion

```

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * Freivald's algorithm
+ * The function will will check AB=C
+ * @param A matrix A
+ * @param B matrix B
+ * @param C resultant matrix C
+ * @returns boolean value
+ */
+template <std::size_t N>
```
Panquesito7, line 24:
```suggestion
/**
 * @namespace others
 * @brief Other algorithms
 */
namespace others {
/**
 * @namespace freivalds
 * @brief Functions for [Freivalds']
 * (https://en.wikipedia.org/wiki/Freivalds%27_algorithm) algorithm
 */
namespace freivalds {
template <std::size_t N>

```

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,163 @@
...
+            flag = false;
+        }
+    }
+    if (flag == true) {
+        std::cout << "\nC is the product of matrix A and C\n";
+    } else {
+        std::cout << "\nC is not the product of matrix A and C\n";
+    }
+    return flag;
+}
```
Panquesito7, line 106:
```suggestion
}
}  // namespace freivald
}  // namespace others

```

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * Error Analysis: If {\displaystyle A\times B=C}{\displaystyle A\times B=C},
+ *then the algorithm always returns "Yes". If {\displaystyle A\times B\neq C}A
+ *\times B \neq C, then the probability that the algorithm returns "Yes" is less
+ *than or equal to one half. This is called one-sided error. By iterating the
+ *algorithm k times and returning "Yes" only if all iterations yield "Yes", a
+ *runtime of O(kn^{2})and error probability of <= 1/(2^k) is achieve
+ */
```
Panquesito7, line 114: I think this shall be provided in the `@details` in the test functions?

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,163 @@
...
+ * @author [Darshana Sarma](https://github.com/Darshana-Sarma)
+ * @details
+ * Given three n × n matrices A, B, and
+ * C, a general problem is to verify whether AB=C
+ *
+ * The time Complexity of the algorithm is O(n^{2}).
+ */
+#include <array>     //for std::array
+#include <cassert>   //for assert
+#include <iostream>  //for io operations
```
Panquesito7, line 14:
```suggestion
#include <array>     //   for std::array
#include <cassert>   //   for assert
#include <iostream>  //   for io operations

```

Panquesito7 (changes requested):

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,176 @@
...
+    std::array<std::array<double, 2>, 2> C = {{{5, 4}, {5, 9}}};
+
+    bool result = others::freivalds::CoppersmithAlgo(A, B, C);
+    std::cout << result << "\n";
+
+    assert(result == false);
+}
+
+/**
+ * Main function
```
Panquesito7, line 169:
```suggestion
 * @brief Main function
 * @returns 0 on exit
```

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,177 @@
...
+ * then the algorithm always returns "Yes". If {\displaystyle A\times B\neq C}A
+ * \times B \neq C, then the probability that the algorithm returns "Yes" is less
+ * than or equal to one half. This is called one-sided error. By iterating the
+ * algorithm k times and returning "Yes" only if all iterations yield "Yes", a
+ * runtime of O(kn^{2})and error probability of <= 1/(2^k) is achieve
+ */
+namespace freivalds {
+template <std::size_t N>
+
+bool CoppersmithAlgo(std::array<std::array<double, N>, N> A,
```
Panquesito7, line 44: Missing function documentation: parameters, brief description and template parameters.

Panquesito7 (changes requested):

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,169 @@
...
+namespace others {
+/**
+ * @namespace freivalds
+ * @brief Functions for [Freivalds']
+ * (https://en.wikipedia.org/wiki/Freivalds%27_algorithm) algorithm
+ */
+namespace freivalds {
+/**
+ * The function will check AB=C by applying Freidvald's algorithm
+ * @param A square matrix A
```
Panquesito7, line 29: Please provide a description for the template parameter.
```suggestion
 * @tparam N description
 * @param A square matrix A
```

Panquesito7 (changes requested):

Panquesito7 (approved):
LGTM. 👍 😄

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * The function will check AB=C by applying Freidvald's algorithm 
+ * @tparam N dimension of square matrix NxN
+ * @param A square matrix A
+ * @param B square matrix B
+ * @param C resultant square matrix C
+ * @returns boolean value
+ */
+template <std::size_t N>
+bool CoppersmithAlgo(std::array<std::array<double, N>, N> A,
```
fhlasek, line 36: pass the input parameters by const reference
```suggestion
bool CoppersmithAlgo(const std::array<std::array<double, N>, N>& A,
```

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,169 @@
+/**
+ * @file
+ * @brief Implementation of [Freivalds' algorithm]
+ * (https://en.wikipedia.org/wiki/Freivalds%27_algorithm)
+ * @author [Darshana Sarma](https://github.com/Darshana-Sarma)
+ * @details
+ * Given three n × n matrices A, B, and
+ * C, a general problem is to verify whether AB=C
+ *
+ * The time Complexity of the algorithm is O(n^{2}).
```
fhlasek, line 10: It'd be worth mentioning that it's a randomised algorithm and what is the probability of success.

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    // Printing matrix C
+    std::cout << "The resultant matrix: \n";
+    for (int i = 0; i < N; i++) {
+        for (int j = 0; j < N; j++) {
+            std::cout << C[i][j] << "\t";
+        }
+        std::cout << "\n";
+    }
+    // generate a random nx1 vector r containing 0/1 elements
```
fhlasek, line 65: If you use the algorithm with a random vector of length `n`, you get a time complexity `O(n^3)` which is as slow as multiplying the matrices altogether (while being a bit worse because it introduces a chance of failure).

I believe the real strength of the algorithm is using it with shorter (constant-sized) random vectors of length k that according to the linked wikipedia article lead to less than `1 / 2^k` probability of failure.

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,169 @@
...
+    std::array<std::array<double, 2>, 2> A = {{{1, 0}, {2, 1}}};
+
+    std::array<std::array<double, 2>, 2> B = {{{5, 4}, {-5, 1}}};
+
+    std::array<std::array<double, 2>, 2> C = {{{5, 4}, {5, 9}}};
+
+    bool result = others::freivalds::CoppersmithAlgo(A, B, C);
+    std::cout << result << "\n";
+
+    assert(result == true);
```
fhlasek, line 138: This test fails when I run it on my computer. There might be some non-determinism.

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,169 @@
...
+    // Printing matrix C
+    std::cout << "The resultant matrix: \n";
+    for (int i = 0; i < N; i++) {
+        for (int j = 0; j < N; j++) {
+            std::cout << C[i][j] << "\t";
+        }
+        std::cout << "\n";
+    }
+    // generate a random nx1 vector r containing 0/1 elements
+    std::array<std::array<const double, N>, 1> a{0};
```
fhlasek, line 66: It seems like a rather bad practice to rely on generating random number from an arbitrary garbage in the memory. There's a good chance you'll come across an empty piece of memory and obtain a vector full of zeros. I suggest you use a proper (pseudo)-random number generator.

fhlasek (changes requested):
I like the algorithm and the implementation! Have a couple suggestions that I believe are needed to make the code really work.

others/freivalds_algorithm.cpp
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+    if (flag == true) {
+        std::cout << "\nC is the product of matrix A and C\n";
+    } else {
+        std::cout << "\nC is not the product of matrix A and C\n";
+    }
+    return flag;
+}
+}  // namespace freivald
```
Panquesito7, line 118:
```suggestion
}  // namespace freivalds

```

others/freivalds_algorithm.cpp
kvedala: I don't think this has anything to do with random number generators.
`a` is simple a vector/array = 1D matrix. 
What I understood was that the initialization `a{0}` will initialize the values of all its elements as 0 as:
1. Initialize the 1st dimension values to 0
2. implicitly applies initialize the elements of `a[0]` with 0.

However, this is counterintuitive to the use of the variable `a` on line 73 and 82.

others/Coppersmith_frievald's.cpp
```diff
@@ -0,0 +1,163 @@
...
+    for (int i = 0; i < N; i++) {
+        for (int j = 0; j < N; j++) {
+            std::cout << C[i][j] << "\t";
+        }
+        std::cout << "\n";
+    }
+    // generate a random nx1 vector r containing 0/1 elements
+    std::array<std::array<const double, N>, 1> a{0};
+
+    // Multiply matrix B with vector 'r' which gives a nx1 vector 'br'
```
ayaankhan98, line 57: instead of writing the multiplication logic again and again it will be better if you create a function
`multiply(mat A, mat B)`
to return the multiplication result, this way code looks more readable and clear and it also reduces the size of code

others/freivalds_algorithm.cpp
fhlasek: From the algorithm description and from the above comment, it is clear that the desired behaviour is to actually produce an array of N random numbers here.

Initialisation by `{0}` is very non-intuitive construct that I think should be avoided completely. If the intention is to have default value initialisation, it can be left out completely with the default constructor. I'm afraid, that according to the [specs](https://en.cppreference.com/w/cpp/container/array) this way when you only initialise the first element, you run into an issue with indeterminate values.

kvedala: Isn't this aggregate initialization: https://en.cppreference.com/w/cpp/language/aggregate_initialization

PR #1232, opened 2020-10-08
fix: Improve Algorithm

akxt1022:
#### Description of Change
* Use ternary operator ```?:``` 
* Using return type ```int64_t``` which will produce correct output for greater numbers.

#### Checklist

Notes: 

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1232"></a>

### DIFF
--- a/math/factorial.cpp
+++ b/math/factorial.cpp
@@ -5,7 +5,7 @@
 #include <iostream>
 
 /** function to find factorial of given number */
-unsigned long factorial(unsigned int n) {
+long factorial(unsigned int n) {
     return n == 0 ? 1 : n * factorial(n - 1);
 }
 
--- a/math/factorial.cpp
+++ b/math/factorial.cpp
@@ -5,7 +5,7 @@
 #include <iostream>
 
 /** function to find factorial of given number */
-long factorial(unsigned int n) {
+int64_t factorial(unsigned int n) {
     return n == 0 ? 1 : n * factorial(n - 1);
 }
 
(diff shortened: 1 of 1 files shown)


math/factorial.cpp
```diff
@@ -5,10 +5,8 @@
 #include <iostream>
 
 /** function to find factorial of given number */
-unsigned int factorial(unsigned int n) {
-    if (n == 0)
-        return 1;
-    return n * factorial(n - 1);
+int64_t factorial(unsigned int n) {
```
kvedala, line 8: the suggested improvement wouldn't really happen until the input argument type is also explicitly specified.
```suggestion
int64_t factorial(uint64_t n) {
```

kvedala (commented):
Though the suggestion is valid (not necessarily an enhancement), it might be better to keep the expanded version in block comments so that a novice reader can still understand.

math/factorial.cpp
akxt1022: Input can be int only because what expected is the output should be correct because output is going to be bigger number. So i think we don't need to change input type.

kvedala: You have a recursive function i.e., the output of the function in one instance is also the input in some next recursion. Hence, the input and the output **must** be of the same datatype. Your tests might be working because the compiler and the machine you are running on might have a default 64-bit or 32-bit integer that is satisfying the condition.

PR #1238, opened 2020-10-10
fix: refactored insertion and heap sort

Groctel:
#### Description of Change

I've refactored the code in the insertion and heap sort to clean them and have them be better self-documented

#### Checklist

Notes: 

Ask away if you require further changes!

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1238"></a>

### DIFF
--- a/sorting/heap_sort.cpp
+++ b/sorting/heap_sort.cpp
@@ -35,8 +35,9 @@
  */
 template <typename T>
 void printArray(T *arr, unsigned size) {
-    for (unsigned i = 0; i < size; i++)
+    for (unsigned i = 0; i < size; i++) {
         std::cout << arr[i] << " ";
+    }
 
     std::cout << std::endl;
 }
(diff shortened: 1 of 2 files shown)


sorting/heap_sort.cpp
```diff
@@ -30,13 +30,16 @@
...
  * @param arr array to be printed
- * @param sz size of array
+ * @param size size of array
  *
  */
 template <typename T>
-void printArray(T *arr, int sz) {
-    for (int i = 0; i < sz; i++) std::cout << arr[i] << "  ";
-    std::cout << "\n";
+void printArray(T *arr, unsigned size) {
```
kvedala, line 37: as long as you are doing the refactoring, it might be better to do it correctly. the correct data type for size of an array and position in an array is not `unsigned` but rather `uint32_t`.

Groctel:
TIL about that specific data type. Thanks! I've fixed it in both files. Also, the style checks fail for barebones arrays and don't let the compile checks start. Shouldn't that be the other way around? :thinking:

kvedala:
> TIL about that specific data type. Thanks! I've fixed it in both files. Also, the style checks fail for barebones arrays and don't let the compile checks start. Shouldn't that be the other way around? 🤔

The workaround would be to use either `std::array` or `std::vector`. There are many codes in the repo using this approach.
[using `std::array`](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/sorting/gnome_sort.cpp)
[using `std::vector`](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/sorting/pancake_sort.cpp)

PR #1260, opened 2020-10-12
remove some bad practice 

x0rld:
there is lot of files with `using namespace std;` who is a bad practice 
I replace the \n by std::endl for better practice
I replace the NULL by nullptr cause it's a C usage not c++

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1260"></a>


xgupta:
@x0rld  best approach is to fix and send pull request for one file at a time. means nine separate pull requests for each file.

x0rld:
ok i close this one

PR #1263, opened 2020-10-12
remove using namespace std;

x0rld:
using namespace std is a bad pratice.

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1263"></a>

### DIFF
--- a/operations_on_datastructures/array_left_rotation.cpp
+++ b/operations_on_datastructures/array_left_rotation.cpp
@@ -1,4 +1,6 @@
 #include <iostream>
+#include <vector>
+
 int main() {
     int n, k;
     std::cout << "Enter size of array=\t";
@@ -7,8 +9,10 @@ int main() {
     std::cin >> k;
     std::vector<int> a;
     std::cout << "Enter  elements of array=\t";
+    int value;
     for (int i = 0; i < n; i++) {
-        std::cin >> a[i];
+        std::cin >> value;
+        a.push_back(value);
     }
     int temp = 0;
     for (int i = 0; i < k; i++) {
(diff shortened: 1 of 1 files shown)


xgupta:
Use std::vector instead of VLA to pass the clang-tidy errors. 
For example replace int n[] to std::vector<int> n;

x0rld:
I will continue later

operations_on_datastructures/array_left_rotation.cpp
```diff
@@ -22,9 +25,9 @@ int main() {
             }
         }
     }
-    cout << "Your rotated array is=\t";
+    std::cout << "Your rotated array is=\t";
     for (int j = 0; j < n; j++) {
-        cout << a[j] << " ";
+        std::cout << a[j] << " ";
     }
     getchar();
```
kvedala, line 32:
```suggestion
```

kvedala (changes requested):
thank you for your contribution

kvedala:
This folder and the code therein is not making much sense to me for this repo. @Panquesito7 @fhlasek @ayaankhan98  thoughts?

fhlasek:
@kvedala I agree with you. On top of it, the code is of such a low quality that I don't think it carries much value. Rather than trying to refactor, I'd suggest to deprecate it and let the community recreate whatever valuable algorithms there might have been following the improved guidelines and best coding practices.

kvedala:
@x0rld Thank you for your contribution. However, the admins find that this folder and the code therein is rudimentary and will need to be either removed altogether or salvaged in to one of the other folders in a more refined code. 

We more than welcome you to contribute in the latter 😄

Issue #1309, opened 2020-10-17
Algorithm to find number of inversions in an array


lordinkavu:
An algorithm to find how many inversions are required to make an array sorted. This would give an idea on the degree to which the algorithm is already sorted.

danglingP0inter:
Can you assign this issue to me?

Ganesh-Agarwal-25:
please assign this to me

aashigupta19:
Sir can you please assign this to me. We are a team of 11 members doing this project as part of our college(LNMIIT) curriculum,  so we'll be completing this project within 4 weeks.

Panquesito7:
> please assign this to me

Please read the [contributing guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#contributor): do **not** ask to be
assigned, instead, submit a PR and maintainers will review it. 🙂

tejasdeore7:
@Panquesito7 I have solved this issue. Please review my PR.

PR #1407, opened 2020-10-28
Fix of issue #1309

tejas702:
Approach :
Traverse through the array and for every index find the number of smaller elements on its right side of the array. This can be done using a nested loop. Sum up the counts for all index in the array and print the sum.
Algorithm : 
    Traverse through the array from start to end
    For every element find the count of elements smaller than the current number upto that index using another loop.
    Sum up the count of inversion for every index.
    Print the count of inversions.

    Complexity Analysis: 
        Time Complexity: O(n^2), Two nested loops are needed to traverse the array from start to end so the Time complexity is O(n^2)
        Space Complexity:O(1), No extra space is required.

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1407"></a>

### DIFF
--- a/sorting/inversions.cpp
+++ b/sorting/inversions.cpp
@@ -2,9 +2,8 @@
 // in an array
 #include <iostream>
 #include <vector>
-using namespace std;
 
-int getInvCount(vector<int>& arr, int n)
+int getInvCount(const std::vector<int>& arr, const int& n)
 {
 	int inv_count = 0;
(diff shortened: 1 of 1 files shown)


Panquesito7 (changes requested):
Code is not up to the [repository standards](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md).
Please read them **carefully** and **follow them**.

PR #962, opened 2020-07-20
feat: Add Vigenère cipher

imdeep2905:
#### Description of Change
Added Vigenère cipher as ciphers/vigenere_cipher.cpp

#### Checklist

- [x] Added description of change
- [x] Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#New-File-Name-guidelines)
- [x] Added tests and example, test must pass
- [x] Added documentation so that the program is self-explanatory and educational - [Doxygen guidelines](https://www.doxygen.nl/manual/docblocks.html)
- [x] Relevant documentation/comments is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#Commit-Guidelines)
- [x] Search previous suggestions before making a new one, as yours may be a duplicate.
- [x] I acknowledge that all my contributions will be made under the project's license.

Notes: 

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/962"></a>

PR #1428, opened 2020-11-09
[fix/docs]: Memory leakage caused by using `new` operator & refactoring/documentation

farbodahm:
#### Checklist

Notes: 
In current version of `data_structures/queue_using_array.cpp` we have a really **bad bug (Resource leakage)** which is caused by using `new` operator without using `delete[]` after that ([Line 24](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/data_structures/queue_using_array.cpp#L24)); I fixed that by using `std::vector`. (If we want to still use C-Style array, just let me know and I will roll back to what we have now but I'll use `Smart Pointer` to fix that bug.)

Also I made some small refactoring to increase quality of code :)

If any thing need to be changed, just let me know that and after that it will be ready :)

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1428"></a>

### DIFF
--- a/data_structures/queue_using_array.cpp
+++ b/data_structures/queue_using_array.cpp
@@ -10,30 +10,31 @@
 #define MAX_SIZE 10
 
 class Queue_Array {
-    int front;
-    int rear;
-
- public:
-    Queue_Array() {
-        front = -1;
-        rear = -1;
-        arr.resize(MAX_SIZE);
-    }
-
-    std::vector<int> arr;
+public:
+    Queue_Array();
     void enqueue(int);
     int dequeue();
-    void display();
+    void display() const;
+private:
+    int front;
+    int rear;
+    std::vector<int> arr;
 };
 
-void Queue_Array::enqueue(int ele) {
+Queue_Array::Queue_Array() {
+    front = -1;
+    rear = -1;
+    arr.resize(MAX_SIZE);
+}
+
+void Queue_Array::enqueue(const int ele) {
     if (rear == arr.size() - 1) {
         std::cout << "\nStack is full";
     } else if (front == -1 && rear == -1) {
         front = rear = 0;
         arr.at(rear) = ele;
     } else if (rear < arr.size()) {
-        rear++;
+        ++rear;
         arr.at(rear) = ele;
     }
 }
@@ -53,11 +54,11 @@ int Queue_Array::dequeue() {
     return d;
 }
 
-void Queue_Array::display() {
+void Queue_Array::display() const {
     if (front == -1) {
         std::cout << "\nStack is empty";
     } else {
-        for (int i = front; i <= rear; i++) std::cout << arr.at(i) << " ";
+        for (int i{front}; i <= rear; ++i) std::cout << arr.at(i) << " ";
     }
 }
 
--- a/data_structures/queue_using_array.cpp
+++ b/data_structures/queue_using_array.cpp
@@ -17,7 +17,7 @@
  * @author [Farbod Ahmadian](https://github.com/farbodahm)
  */
 #include <iostream> /// for io operations
-#include <vector>   /// for std::vector
+#include <array>   /// for std::array
 
 constexpr uint16_t max_size{10}; ///< Maximum size of the queue
 
@@ -39,37 +39,29 @@ namespace queue_using_array {
  */
 class Queue_Array {
 public:
-    Queue_Array();                ///< Set maximum number of data 
-    void enqueue(const int8_t&);  ///< Add element to the first of the queue
+    void enqueue(const int16_t&);  ///< Add element to the first of the queue
     int dequeue();                ///< Delete element from back of the queue
     void display() const;         ///< Show all saved data
 private:
     int8_t front{-1};             ///< Index of head of the array
     int8_t rear{-1};              ///< Index of tail of the array
-    std::vector<int8_t> arr{};    ///< All stored data
+    std::array<int16_t, max_size> arr;    ///< All stored data
 };
-
-
-/**
-* @brief Queue_Array constructor. Initializes the maximum number of data than can be saved.
-*/
-Queue_Array::Queue_Array() {
-    arr.resize(max_size);
-}
 
 /**
  * @brief Adds new element to the end of the queue
  * @param ele to be added to the end of the queue
  */
-void Queue_Array::enqueue(const int8_t& ele) {
+void Queue_Array::enqueue(const int16_t& ele ) {
     if (rear == arr.size() - 1) {
         std::cout << "\nStack is full";
     } else if (front == -1 && rear == -1) {
-        front = rear = 0;
-        arr.at(rear) = ele;
+        front = 0;
+        rear = 0;
+        arr[rear] = ele;
     } else if (rear < arr.size()) {
         ++rear;
-        arr.at(rear) = ele;
+        arr[rear] = ele;
     }
 }
 
@@ -99,7 +91,7 @@ void Queue_Array::display() const {
     if (front == -1) {
         std::cout << "\nStack is empty";
     } else {
-        for (int8_t i{front}; i <= rear; ++i) std::cout << arr.at(i) << " ";
+        for (int16_t i{front}; i <= rear; ++i) std::cout << arr.at(i) << " ";
     }
 }
 
--- a/data_structures/queue_using_array.cpp
+++ b/data_structures/queue_using_array.cpp
@@ -1,14 +1,37 @@
-/*
-    Write a program to implement Linear Queue using array.
-    Functions to implement
-        Enqueue (Insertion)
-        Dequeue (Deletion)
-*/
+/**
+ * @file
+ * @brief Implementation of Linear Queue using array.
+ * @details
+ * The Linear Queue is a data structure used for holding a sequence of
+ * values, which can be added to the end line (enqueue), removed from
+ * head of line (dequeue) and displayed.
+ * ### Algorithm
+ * Values can be added by increasing the `rear` variable by 1 (which points to 
+ * the end of the array), then assigning new value to `rear`'s element of the array.
+ * 
+ * Values can be removed by increasing the `front` variable by 1 (which points to 
+ * the first of the array), so it cannot reached any more.
+ */
 #include <iostream>
 #include <vector>
 
 constexpr uint8_t max_size{10};
 
+/**
+ * @namespace data_structures
+ * @brief Data Structures algorithms
+ */
+namespace data_structures {
+
+/**
+ * @namespace linked_list
+ * @brief Functions for singly linked list algorithm
+ */
+namespace linked_list {
+
+/**
+ * Queue_Array class containing the main data and also index of head and tail of the array.
+ */
 class Queue_Array {
 public:
     Queue_Array();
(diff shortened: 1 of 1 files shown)


data_structures/queue_using_array.cpp
```diff
@@ -1,69 +1,69 @@
...
+public:
+    Queue_Array();
     void enqueue(int);
     int dequeue();
-    void display();
+    void display() const;
+private:
+    int front;
+    int rear;
+    std::vector<int> arr;
```
Panquesito7, line 21:
```suggestion
    int front = 0;
    int rear = 0;
    std::vector<int> arr{};
```

data_structures/queue_using_array.cpp
```diff
@@ -1,69 +1,69 @@
...
 
-void Queue_Array::enqueue(int ele) {
-    if (rear == size - 1) {
+Queue_Array::Queue_Array() {
+    front = -1;
+    rear = -1;
+    arr.resize(max_size);
+}
+
+void Queue_Array::enqueue(const int ele) {
```
Panquesito7, line 30:
```suggestion
void Queue_Array::enqueue(const int& ele) {
```

data_structures/queue_using_array.cpp
```diff
@@ -1,69 +1,69 @@
...
-    }
-    int *arr = new int[size];
+public:
+    Queue_Array();
     void enqueue(int);
     int dequeue();
-    void display();
+    void display() const;
+private:
+    int front;
```
Panquesito7, line 19: It is recommended to use `uint64_t` for non-negative values or `int64_t` for negative values (or their appropriate size). Check other parts of the code ([reference](https://en.cppreference.com/w/cpp/types/integer)).

Panquesito7 (commented):
👍 good work.

data_structures/queue_using_array.cpp
farbodahm: Dear @Panquesito7 ,
I changed them as you said in upcoming commit, but I have a question for my curiosity:
Should we initialize private member variables when we declare them or in class's constructor? Can I know why for further information? :D
Thanks for your help

data_structures/queue_using_array.cpp
farbodahm: Thank you,
What will happen if we use larger size for most cases (`uint64_t`), Will it waste memory?

farbodahm:
Dear @Panquesito7 ,
Thanks for your quick and great review, requested changed are applied.

data_structures/queue_using_array.cpp
Panquesito7: Yes, wasting memory is not good. It could lead to memory overflow, too.

data_structures/queue_using_array.cpp
Panquesito7: > when we declare them or in class's constructor?

I think it's better to initialize them first when you declare them, and you can
then re-initialize them in the class constructor. It should not harm the code.

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,75 @@
...
 
     std::cout << "\n1. enqueue(Insertion) ";
     std::cout << "\n2. dequeue(Deletion)";
     std::cout << "\n3. Display";
     std::cout << "\n4. Exit";
-    while (1) {
+    while (true) {
         std::cout << "\nEnter your choice ";
         std::cin >> op;
         if (op == 1) {
```
Panquesito7, line 75: Please add `return 0;` at the end of `main()` function.

farbodahm:
Sorry for delayed update, it was late night in my timezone : )
Pr is updated.

Panquesito7 (approved):
Good work, thank you for your contribution. 👍 🙂
Now, this code is a bit more refined, and self-tests are passing too.

We could take the opportunity to document the code and make it educational (in another PR).
Confused, or need a reference? See #962 and the [typical structure](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#typical-structure-of-a-program) of a program. 🙂

farbodahm:
Code is now documented.
@Panquesito7 Would you mind please check that?

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,122 @@
...
+/**
+ * @namespace data_structures
+ * @brief Data Structures algorithms
+ */
+namespace data_structures {
 
-#define MAXSIZE 10
+/**
+ * @namespace linked_list
+ * @brief Functions for singly linked list algorithm
```
Panquesito7, line 28:
```suggestion
 * @brief Functions for [Queue using Array](https://www.geeksforgeeks.org/array-implementation-of-queue-simple/) implementation
```

Panquesito7 (changes requested):
👍 Well done!

farbodahm:
Doc is now updated, just let me know if any other changes are required.

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,127 @@
...
-        size = MAXSIZE;
-    }
-    int *arr = new int[size];
-    void enqueue(int);
+public:
+    Queue_Array();
+    void enqueue(const int8_t&);
     int dequeue();
-    void display();
+    void display() const;
```
Panquesito7, line 45: Add a brief description of what these functions do (see how private variables are documented).

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,127 @@
...
+
+/**
+* Queue_Array constructor. Initializes the maximum number of data than can be saved.
+*/
+Queue_Array::Queue_Array() {
+    arr.resize(max_size);
+}
+
+/**
+ * function adds new element to the end of the queue
```
Panquesito7, line 61:
```suggestion
 * @brief Adds new element to the end of the queue
```

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,127 @@
...
-        arr[rear] = ele;
+        arr.at(rear) = ele;
+    } else if (rear < arr.size()) {
+        ++rear;
+        arr.at(rear) = ele;
     }
 }
 
+/**
+ * function removes the element at the first of the queue
```
Panquesito7, line 77:
```suggestion
 * @brief Removes the element at the first of the queue
```

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,127 @@
...
-        d = arr[front++];
+        d = arr.at(front++);
     }
 
     return d;
 }
 
-void Queue_Array::display() {
+/**
+ * function shows all elements in the queue
```
Panquesito7, line 96:
```suggestion
 * @brief Utility function to show all elements in the queue
```

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,127 @@
...
         std::cout << "\nStack is empty";
     } else {
-        for (int i = front; i <= rear; i++) std::cout << arr[i] << " ";
+        for (int8_t i{front}; i <= rear; ++i) std::cout << arr.at(i) << " ";
     }
 }
 
-int main() {
-    int op, data;
+}  // namespace linked_list
```
Panquesito7, line 106:
```suggestion
}  // namespace queue_using_array
```

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,127 @@
...
+}  // namespace data_structures
 
-    Queue_Array ob;
+
+/**
+ * Main function:
```
Panquesito7, line 111:
```suggestion
 * @brief Main function
```

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,127 @@
...
+    int8_t front{-1};           ///< Index of head of the array
+    int8_t rear{-1};            ///< Index of tail of the array
+    std::vector<int8_t> arr{};  ///< All stored data
 };
 
-void Queue_Array::enqueue(int ele) {
-    if (rear == size - 1) {
+
+/**
+* Queue_Array constructor. Initializes the maximum number of data than can be saved.
```
Panquesito7, line 54:
```suggestion
* @brief Queue_Array constructor. Initializes the maximum number of data than can be saved.
```

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,127 @@
...
-int main() {
-    int op, data;
+}  // namespace linked_list
+}  // namespace data_structures
 
-    Queue_Array ob;
+
+/**
+ * Main function:
+ * Allows the user add and delete values from the queue.
```
Panquesito7, line 112:
```suggestion
 *
 * @details
 * Allows the user add and delete values from the queue.
```

Panquesito7 (changes requested):

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,128 @@
...
+ * @author [Pooja](https://github.com/pooja-git11)
+ * @author [Farbod Ahmadian](https://github.com/farbodahm)
+ */
+#include <iostream> /// for io operations
+#include <vector>   /// for std::vector
 
-    Functions to implement
-        Enqueue (Insertion)
-        Dequeue (Deletion)
+constexpr uint8_t max_size{10};
```
Panquesito7, line 22: Add a brief description of what this variable does (see how private class variables are documented).

farbodahm:
Thanks for your time, changes applied.

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,128 @@
...
+ * @author [Pooja](https://github.com/pooja-git11)
+ * @author [Farbod Ahmadian](https://github.com/farbodahm)
+ */
+#include <iostream> /// for io operations
+#include <vector>   /// for std::vector
 
-    Functions to implement
-        Enqueue (Insertion)
-        Dequeue (Deletion)
+constexpr uint8_t max_size{10}; ///< Maximum size of the queue
```
Panquesito7, line 22: I think this should be `uint16_t`, to prevent overflows
```suggestion
constexpr uint16_t max_size{10}; ///< Maximum size of the queue
```

Panquesito7 (approved):
Awesome work! Thanks for your contribution! :tada: 👍

data_structures/queue_using_array.cpp
```diff
@@ -1,77 +1,128 @@
...
+ * @author [Pooja](https://github.com/pooja-git11)
+ * @author [Farbod Ahmadian](https://github.com/farbodahm)
+ */
+#include <iostream> /// for io operations
+#include <vector>   /// for std::vector
 
-    Functions to implement
-        Enqueue (Insertion)
-        Dequeue (Deletion)
+constexpr uint16_t max_size{10}; ///< Maximum size of the queue
```
ayaankhan98, line 22: @farbodahm Great improvement over documentation and code fixes, Thanks for your contribution.
Here is a suggestion, basically vectors are used to store the queue elements so there is no need of queue size, it would be great if you do a little imrpovement over this.

farbodahm: Thanks for your great review,

If we don't use a variable to store `max size` and just rely on `std::vector`, then this code will not be an implementation for **linear** queue because it can grow as much as user want.
So we can use `std::array` instead of `std::vector` to have `linear` queue OR we can still use `std::vector` and delete `max_size` variable but it will not be a linear queue any more.

Just let me know which one is preferred, and I will change that as you wished :)

ayaankhan98: i think it's good to use arrays because the idea behind the implementation of these data structures is to give the foundational idea of how the standard data structures like `std::list` or `std::deque` are implemented on top of basic arrays.
And as here we are dealing with C++ it's good to use `std::array` instead of raw C style arrays.
Hope this makes sense.

farbodahm:
@ayaankhan98 Thank you, now we are using `std::array` instead of `std::vector`; just as you said, it makes sense now.

ayaankhan98 (approved):
LGTM! Thanks @farbodahm

PR #1429, opened 2020-11-15
Feat: added n_choose_r combinatorics algorithm

tjgurwara99:
#### Description of Change

Added a simple n choose r combinatoric function.

#### Checklist

Notes:  Added a simple n choose r combinatoric function.

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1429"></a>

### DIFF
--- a/math/n_choose_r.cpp
+++ b/math/n_choose_r.cpp
@@ -1,6 +1,6 @@
 /**
  * @file
- * @brief Combinations n choose r function implementation
+ * @brief [Combinations](https://en.wikipedia.org/wiki/Combination) n choose r function implementation
  * @details
  * A very basic and efficient method of calculating
  * choosing r from n different choices. For more information visit 
--- a/math/n_choose_r.cpp
+++ b/math/n_choose_r.cpp
@@ -4,7 +4,7 @@
  * @details
  * A very basic and efficient method of calculating
  * choosing r from n different choices.
- * $ \binom{n}{r} = \frac{n!}{r! (n-r)!} $
+ * \f$ \binom{n}{r} = \frac{n!}{r! (n-r)!} \f$
  *
  * @author [Taj](https://github.com/tjgurwara99)
  */
@@ -13,15 +13,15 @@
 #include <cassert>
 
 /**
- * This is the function implementation of $\binom{n}{r}$
+ * This is the function implementation of \f$ \binom{n}{r} \f$
  * We are calculating the ans with iterations
  * instead of calculating three different factorials.
  * Also, we are using the fact that 
- * $ \frac{n!}{r! (n-r)!} = \frac{(n - r + 1) \cdot \cdots \cdot n}{1 \cdot \cdots \cdot r} $
+ * \f$ \frac{n!}{r! (n-r)!} = \frac{(n - r + 1) \cdot \cdots \cdot n}{1 \cdot \cdots \cdot r} \f$
  * @tparam T Only for integer types such as long, int_64 etc
- * @param n $n$ in $ \binom{n}{r} $
- * @param r $r$ in $ \binom{n}{r} $
- * @returns ans $\binom{n}{r} $
+ * @param n \f$ n \f$ in \f$ \binom{n}{r} \f$
+ * @param r \f$ r \f$ in \f$ \binom{n}{r} \f$
+ * @returns ans \f$ \binom{n}{r} \f$
  */
 template <class T>
 T n_choose_r(T n, T r) {
--- a/math/n_choose_r.cpp
+++ b/math/n_choose_r.cpp
@@ -3,8 +3,7 @@
  * @brief [Combinations](https://en.wikipedia.org/wiki/Combination) n choose r function implementation
  * @details
  * A very basic and efficient method of calculating
- * choosing r from n different choices. For more information visit 
- * [Combination](https://en.wikipedia.org/wiki/Combination) on Wikipedia.
+ * choosing r from n different choices.
  * \f$ \binom{n}{r} = \frac{n!}{r! (n-r)!} \f$
  *
  * @author [Tajmeet Singh](https://github.com/tjgurwara99)
--- a/math/n_choose_r.cpp
+++ b/math/n_choose_r.cpp
@@ -42,22 +42,22 @@ T n_choose_r(T n, T r) {
  */
 static void test() {
     // First test on 5 choose 2
-     long t = (long) n_choose_r(5, 2);
+     int t = n_choose_r(5, 2);
      assert(((void)"10 is the answer but function says otherwise.\n",
              t == 10));
      std::cout << "First test passes." << std::endl;
      // Second test on 5 choose 3
-     t = (long) n_choose_r(5, 3);
+     t = n_choose_r(5, 3);
      assert(((void)"10 is the answer but the function says otherwise.\n",
              t == 10));
      std::cout << "Second test passes." << std::endl;
      // Third test on 3 choose 2
-     t = (long)  n_choose_r(3, 2);
+     t = n_choose_r(3, 2);
      assert(((void)"3 is the answer but the function says otherwise.\n",
              t == 3));
      std::cout << "Third test passes." << std::endl;
      // Fourth test on 10 choose 4
-     t = (long) n_choose_r(10, 4);
+     t = n_choose_r(10, 4);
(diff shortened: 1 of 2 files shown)


math/n_choose_r.cpp
```diff
@@ -0,0 +1,76 @@
...
+     assert(((void)"10 is the answer but function says otherwise.\n",
+             t == 10));
+     std::cout << "First test passes." << std::endl;
+     // Second test on 5 choose 3
+     t = n_choose_r(5, 3);
+     assert(((void)"10 is the answer but the function says otherwise.\n",
+             t == 10));
+     std::cout << "Second test passes." << std::endl;
+     // Third test on 3 choose 2
+     t = n_choose_r(3, 2);
```
Panquesito7, line 55:
```suggestion
     t = math::n_choose_r(3, 2);
```

math/n_choose_r.cpp
```diff
@@ -0,0 +1,76 @@
...
+     assert(((void)"3 is the answer but the function says otherwise.\n",
+             t == 3));
+     std::cout << "Third test passes." << std::endl;
+     // Fourth test on 10 choose 4
+     t = n_choose_r(10, 4);
```
Panquesito7, line 60:
```suggestion
     t = math::n_choose_r(10, 4);
```

Panquesito7 (changes requested):
👍 well done!
Please enable [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions) in your repository of this fork in this link: https://github.com/tjgurwara99/C-Plus-Plus/actions

math/n_choose_r.cpp
```diff
@@ -0,0 +1,76 @@
+/**
+ * @file
+ * @brief Combinations n choose r function implementation
```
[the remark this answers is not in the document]
Panquesito7: Not resolved yet (do not resolve conversations that are **not** resolved).

math/n_choose_r.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <iostream>   /// for io operations
+#include <cassert>   /// for assert
+
+
+/**
+ * @namespace math
+ * @brief Mathematical algorithms
+ */
+namespace math {
+
```
Panquesito7, line 22:
```suggestion
```

math/n_choose_r.cpp
```diff
@@ -0,0 +1,84 @@
...
+    if(r > n / 2)
+        r = n - r; // Because of the fact that  nCr(n, r) == nCr(n, n - r)
+    T ans = 1;
+    for(int i = 1; i <= r; i++) {
+        ans *= n - r + i;
+        ans /= i;
+    }
+    return ans;
+}
+} // namespace math closing
```
Panquesito7, line 46:
```suggestion
} // namespace math
```

math/n_choose_r.cpp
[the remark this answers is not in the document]
tjgurwara99: I added the link at the bottom of the explanation. Do you want me to add it at the top?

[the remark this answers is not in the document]
tjgurwara99: I kept the bottom part and added another link to the `@brief` if that was what you wanted. Do let me know if there is something else that needs to be changed.

math/n_choose_r.cpp
```diff
@@ -0,0 +1,83 @@
+/**
+ * @file
+ * @brief [Combinations](https://en.wikipedia.org/wiki/Combination) n choose r function implementation
+ * @details
+ * A very basic and efficient method of calculating
+ * choosing r from n different choices. For more information visit 
+ * [Combination](https://en.wikipedia.org/wiki/Combination) on Wikipedia.
```
Panquesito7, line 7: No need to add it here once it is on the brief description.
```suggestion
```

tjgurwara99: Okay, I'll change it now

math/n_choose_r.cpp
```diff
@@ -0,0 +1,82 @@
...
+ * choosing r from n different choices.
+ * \f$ \binom{n}{r} = \frac{n!}{r! (n-r)!} \f$
+ *
+ * @author [Tajmeet Singh](https://github.com/tjgurwara99)
+ */
+
+#include <iostream>   /// for io operations
+#include <cassert>   /// for assert
+
+
```
Panquesito7, line 15:
```suggestion
```

math/n_choose_r.cpp
```diff
@@ -0,0 +1,82 @@
...
+T n_choose_r(T n, T r) {
+    if(r > n / 2)
+        r = n - r; // Because of the fact that  nCr(n, r) == nCr(n, n - r)
+    T ans = 1;
+    for(int i = 1; i <= r; i++) {
+        ans *= n - r + i;
+        ans /= i;
+    }
+    return ans;
+}
```
ayaankhan98, line 43: I think it makes more sense to use template specilization in this case, because that will automatically ensure that parameters are nothing other than `int` or `int64_t` or `long` whereas in this implementation there will not be any error if someone passes parameters of type `double` or something else rather it will produce wrong results.

tjgurwara99: I think specialisation wouldn't make sense in this case because we don't want a different behaviour for other datatype but rather we want them to nothing if its anything other than integer types - overloading would make more sense. If I were to specialise this function it would very much look like the following (this is the only way I can think of but if you have suggestions, please let me know):

```C++
template <class T>
T n_choose_r(T n, T r) {
    // this should do nothing because we want to specialise to int types
}

template <>
int_64 n_choose_r(int_64 n, int_64 r) {
    if(r > n / 2)
        r = n - r; // Because of the fact that  nCr(n, r) == nCr(n, n - r)
    int_64 ans = 1;
    for(int i = 1; i <= r; i++) {
        ans *= n - r + i;
        ans /= i;
    }
    return ans;
}


```
and so on for other types which makes the idea of having templates redundant (unless you have a better way - if so please let me know as it would help me understand specialisation better). Also consider the case that we have a custom BigInt library and addition, subtraction, multiplication & division is defined for them then this would not work.

Let me know what you think about it😄 I will try to implement what you think would be best.

ayaankhan98: yes you are right using template specilization leads to redundant code. A better way to go with this is to use some sort of functionality that will `ON` or `OFF`  the single general template based on some sort of condition (condition is based on the types of parameters passed in this case) and to achieve this we can have two ways, one is `tag dispatch` and other is `SFINAE`. You can use any one of them.

ayaankhan98: But i think this would make this simple program quite complex for beginners to understand. isn't it?

tjgurwara99: I quite agree, it's meant to be a very simple yet efficient way of calculating nCk and adding both `tag dispatch` and `substitution failure not an error` would indeed just make it very complicated to understand for beginners. If TheAlgorithms collaborative is working on a library then it may be useful for converting this into a STL like function using `tag dispatcher` but if not then I don't think we should do anything to it. What do you think @ayaankhan98?

ayaankhan98: yes the soul purpose of these implementaions is to provide a educational medium so there is no need to make simple things tough, better to leave it as it is.

ayaankhan98: @tjgurwara99 Thanks for your contribution

tjgurwara99: Also, do you have any idea why the `gh-pages` for my fork of this repo is having trouble with graphviz? for example, if you go to the [tjgurwara99.github.io/C-Plus-Plus/n_choose_r](https://tjgurwara99.github.io/C-Plus-Plus/d1/dbb/n__choose__r_8cpp.html#details) you would see that there is an error on it. Any idea of what that is about? 

I ask this because when I compile this same code in the gitpod above it just gives me correct html file with no errors so was just wondering if I have to change something in the Doxygen workflow or something.

ayaankhan98 (approved):
LGTM!

tjgurwara99:
> LGTM!

Has LGTM generated any warnings for this PR?

Panquesito7:
> > LGTM!
> 
> Has LGTM generated any warnings for this PR?

No, there are no warnings.

Panquesito7 (approved):
Great work, @tjgurwara99. Thanks for your contribution! 😄 👍

PR #1530, opened 2021-07-08
Replace "\n" -> std::endl

Jxtopher:
Replace "\n" -> std::endl

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1530"></a>

### DIFF
--- a/backtracking/graph_coloring.cpp
+++ b/backtracking/graph_coloring.cpp
@@ -30,11 +30,11 @@ namespace backtracking {
      */
     template <size_t V>
     void printSolution(const std::array <int, V>& color) {
-        std::cout << "Following are the assigned colors\n";
+        std::cout << "Following are the assigned colors" << std::endl;
         for (auto &col : color) {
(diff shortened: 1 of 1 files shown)


Panquesito7:
Hmm, which one do you think is better: `\n` or `std::endl`? @ayaankhan98, @mishraabhinn.

mishraabhinn:
> Hmm, which one do you think is better: `\n` or `std::endl`? @ayaankhan98, @mishraabhinn.

Yes. We can use `std::endl`, the only difference it also flush the buffer.

Jxtopher:
See: SL.io.50 (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rio-endl)

"Apart from the (occasionally important) issue of performance, the choice between '\n' and endl is almost completely aesthetic."

Panquesito7 (approved):
Thanks!

PR #1605, opened 2021-09-13
Add working workflow for stale items 

siriak:
Stale bot doesn't work as expected, so I moved this responsibility to Stale workflow

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1605"></a>


.github/workflows/stale.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '0 0 * * *'
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4
```
Panquesito7, line 9: I prefer to use the latest version of it.
```suggestion
      - uses: actions/stale@main
```

.github/workflows/stale.yml
```diff
@@ -0,0 +1,19 @@
...
+        with:
+          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
+          close-issue-message: 'This issue was closed because it has been stalled for 7 days with no activity.'
+          stale-pr-message: 'This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
+          close-pr-message: 'This PR was closed because it has been stalled for 7 days with no activity.'
```
Panquesito7, line 14: I would like to keep the messages from the other file. 🙂

Panquesito7:
Thank you for making this pull request! I remember we were going to use this before, but we chose the current one we have for some reason. @ayaankhan98, @mishraabhinn, which one do you think we should use?

siriak:
The current one doesn't work consistently and there are many people complaining about it on the internet (e.g. https://github.com/probot/stale/issues/349)

Panquesito7 (approved):
LGTM 🚀 Thank you for improving the stale bot! 😄👍

PR #1607, opened 2021-09-14
feat: Unleash full power of the stale workflow

siriak:
Set stale bot action limit to default because it's actually GitHub API action limit and not the number of issues/PRs it marks per run as I thought

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1607"></a>


Panquesito7 (approved):
Great catch, thanks! 😄

PR #1809, opened 2021-10-28
[Feat] tuple type

jay-tux:
#### Description of Change
Added a (naive) implementation of C++'s `std::tuple`.  
However, some of the things required (`if constexpr`, `auto` return type) require C++17.

#### Checklist

Notes: requires C++17 (hence the version bump in the master CMakeLists.txt); however, this didn't break anything (`cmake .. && cmake --build .` in a `/build/` succeeded)

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1809"></a>


CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ project(Algorithms_in_C++
 # set(CMAKE_CXX_CPPLINT "~/anaconda3/bin/cpplint --filter=-legal/copyright --std=c++11")
 # find_program(CLANG_FORMAT "clang-format")
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
```
Panquesito7, line 11: Wouldn't we have to update all the other codes to match the C++-17 syntax?

jay-tux: I changed the general CMake file; I'll check any other now and update them all

jay-tux: Seems there are no others which define their own C++ version.  
Compilation threw a warning from another algorithm, yet it seems that this should've existed before (no return in non-void)

Panquesito7: So what should we do, then? I'm confused. 🤔

Panquesito7: > I think the issue is in the CI: could it be that it is set to use C++11?

Hmm, I think it's due to that, not sure though. Maybe we should try to make it work in C++11.

PR #1899, opened 2022-01-30
add algorithm for finding the length of longest substring with no rep…

tazyqui:
I added an algorithm for this problem that uses the sliding window technique and an array instead of a set. This optimizes both the speed and space complexity of the set implementation.

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus/pull/1899"></a>


tazyqui:
This is also my first contribution so please let me know if there is anything I need to do

PR #1998, opened 2022-09-24
Update linear_search.cpp

shrutikahilale:
#### Description of Change

- Improve code regarding the array size input entered by user.

#### Checklist

Notes:

### DIFF
--- a/search/linear_search.cpp
+++ b/search/linear_search.cpp
@@ -82,18 +82,10 @@ int main()
         std::cout << "\nEnter the size of the array [in range 1-30 ]: ";
         std::cin >> size;
 
-        while (size <= 0 || size > 30)
-        {
-            if (size <= 0)
-            {
-                std::cout << "Size cannot be less than zero. Please choose another value: ";
-                std::cin >> size;
-            }
-            if (size > 30)
-            {
-                std::cout << "Size entered more than 30. Please choose a lower value: ";
-                std::cin >> size;
-            }
+        while (size <= 0 || size > 30){
+            std::cout << "Size can only be 1-30. Please choose another value: ";
+            std::cin >> size;
+        }
         }
 
         int *array = new int[size];
--- a/search/linear_search.cpp
+++ b/search/linear_search.cpp
@@ -7,8 +7,8 @@
  * @author [Ritika Mukherjee](https://github.com/ritikaa17)
  */
 
-#include <iostream>   /// for IO operations
-#include <cassert>    /// for assert
+#include <iostream> /// for IO operations
+#include <cassert>  /// for assert
 
 /**
  * \brief [Algorithm implementation for linear search]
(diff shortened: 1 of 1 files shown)


search/linear_search.cpp
```diff
@@ -63,30 +67,42 @@ static void tests() {
...
+    {
         int size = 0;
-        std::cout << "\nEnter the size of the array: ";
+        std::cout << "\nEnter the size of the array [in range 1-30 ]: ";
         std::cin >> size;
 
-        while ((size <= 1) || (size >= 30)) {
-            std::cout << "Size cannot be less than zero. Please choose another value: ";
-            std::cin >> size;
+        while (size <= 0 || size > 30)
```
Panquesito7, line 85: Hmm, I'm not sure what's the difference here. Could you explain the benefits, please?

shrutikahilale: Previously user was prompted to enter array size again even if it was greater than 0 (but it prompted "size cannot be less than 0"), because the user may have entered the size greater than 30. So, for the "user's clarity" regarding the input, the prompt should be made accordingly.

search/linear_search.cpp
```diff
@@ -63,30 +67,42 @@ static void tests() {
...
+            if (size <= 0)
+            {
+                std::cout << "Size cannot be less than zero. Please choose another value: ";
+                std::cin >> size;
+            }
+            if (size > 30)
+            {
+                std::cout << "Size entered more than 30. Please choose a lower value: ";
+                std::cin >> size;
+            }
```
mishraabhinn, line 96: I really feel code getting brittle. Can not we rethink? And we have to take input anyways so input can be taken anyways should not be in if block.

mishraabhinn (changes requested):

search/linear_search.cpp
shrutikahilale: The second if condition is really not needed, but for the programmer's concern of taking size less than 30 that part of code has been added.

Panquesito7, line 96: To make it shorter, we could do something like this. What do you think? 🙂
```suggestion
        while (size <= 0 || size > 30){
            std::cout << "Size can only be 1-30. Please choose another value: ";
            std::cin >> size;
        }
```

shrutikahilale: Yes this looks fine, thankyou 🙂
Should I commit this suggestion?

mishraabhinn: Looks good guys.

mishraabhinn (approved):

search/linear_search.cpp
```diff
@@ -63,30 +67,34 @@ static void tests() {
...
+        std::cout << "\nEnter the size of the array [in range 1-30 ]: ";
         std::cin >> size;
 
-        while ((size <= 1) || (size >= 30)) {
-            std::cout << "Size cannot be less than zero. Please choose another value: ";
+        while (size <= 0 || size > 30){
+            std::cout << "Size can only be 1-30. Please choose another value: ";
             std::cin >> size;
         }
+        }
```
Panquesito7, line 89: This seems to be causing the CI to fail.
```suggestion
```

Panquesito7 (changes requested):

Panquesito7 (approved):
Looks great. Thanks! 🚀

PR #2012, opened 2022-10-01
Create perfect_number.cpp

Anubhav-Ghosh1:
Added code to find perfect number

### DIFF
--- a/math/perfect_number.cpp
+++ b/math/perfect_number.cpp
@@ -1,0 +1,35 @@
+#include <iostream> 
+#include <cctype>
+using namespace std; 
+
+int main(){
+    int n,i=1,sum=0;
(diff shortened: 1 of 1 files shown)


math/perfect_number.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+  1 + 2+ 3 = 6
+
+  and 28 is also a Perfect Number
+
+  since 1+ 2 + 4 + 7 + 14= 28
+
+  Other perfect numbers: 496, 8128
+
+*/
```
mishraabhinn, line 35: Please read contributing guidelines. Such type of code is not acceptable.

mishraabhinn (changes requested):

PR #2058, opened 2022-10-02
Add namespace std

H-e-m-a-n-t:
Removed unnecessary clutter in the code by adding namespace std

#### Checklist

Notes:


mdfaizanahmed786 (commented):
Importing the entire namespace isn't always the right thing to do. For smaller programs, it is sufficient to use a scope resolution operator. We can also use other alternative options such as using directive or typedef operator as well.

Panquesito7:
Sorry, but these changes do not meet our standards. Thank you for contributing.

PR #2109, opened 2022-10-05
Finding no of set bits using existing inbuild library

deoliaman00:
To find the number of set bits in an efficient way. It consumes less time complexity than the previously given solution O(logn) solution.

### DIFF
--- a/bit_manipulation/count_of_set_bits2.cpp
+++ b/bit_manipulation/count_of_set_bits2.cpp
@@ -1,0 +1,69 @@
+/**
+ * @file
+ * @brief Finding No of Set Bits in O(k).
+ * https://www.geeksforgeeks.org/builtin-functions-gcc-compiler/
+ * @details
+ * How to calculate the no of Set bits in a number.
+ * math equations, etc.
+ * @author [Aman Deoli](https://github.com/deoliaman00)
+ * @see related_file.cpp, another_file.cpp
+ */
+
+#include <cassert>   /// for assert
+#include <stdio.h>   //stdio.h is a header file which has the necessary information to include the input/output related functions in our program.
+#include <iostream>
+/**
+ * @namespace bit_manipulation
+ * @brief Bit manipulation algorithms
+ */
+namespace bit_manipulation {
+/**
+ * @namespace no_of_set_bits
+ * @brief Functions for the [count sets
+ * bits](https://www.geeksforgeeks.org/builtin-functions-gcc-compiler/)
+ * implementation
+ */
+namespace no_of_set_bits {
+/**
+ * @brief The main function implements set bit count
+ * @param n is the number whose set bit will be counted
+ * @returns total number of set-bits in the binary representation of number `n`
+ */
+std::uint64_t countSetBits(
+    std ::int64_t n) {  // int64_t is preferred over int so that
+                        // no Overflow can be there.
+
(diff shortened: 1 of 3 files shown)


bit_manipulation/count_of_set_bits2.cpp
```diff
@@ -0,0 +1,69 @@
...
+ * @details
+ * How to calculate the no of Set bits in a number.
+ * math equations, etc.
+ * @author [Aman Deoli](https://github.com/deoliaman00)
+ * @see related_file.cpp, another_file.cpp
+ */
+
+#include <cassert>   /// for assert
+#include <stdio.h>   //stdio.h is a header file which has the necessary information to include the input/output related functions in our program.
+#include <iostream>
```
Panquesito7, line 14: Use `iostream` only. `stdio.h` is used only in C.
```suggestion
#include <iostream>    /// for IO operations
```

bit_manipulation/count_of_set_bits2.cpp
```diff
@@ -0,0 +1,69 @@
...
+                        // no Overflow can be there.
+
+    return __builtin_popcount(n);
+    // Why this algorithm is better than the standard one?
+    // Because this algorithm takes O(k) time where k is no of set bits
+}
+}  // namespace no_of_set_bits
+}  // namespace bit_manipulation
+
+static void test() {
```
Panquesito7, line 43:
```suggestion
/**
 * @brief Self-test implementations
 * @returns void
 */
static void test() {
```

bit_manipulation/count_of_set_bits2.cpp
```diff
@@ -0,0 +1,69 @@
...
+    // n = 15 return 4
+    assert(bit_manipulation::no_of_set_bits::countSetBits(15) == 4);
+    // n = 25 return 3
+    assert(bit_manipulation::no_of_set_bits::countSetBits(25) == 3);
+    // n = 31 return 5
+    assert(bit_manipulation::no_of_set_bits::countSetBits(31) == 5);
+    // n = 32 return 1
+    assert(bit_manipulation::no_of_set_bits::countSetBits(32) == 1);
+    std::cout << "Test cases passed successfully" << std::endl;
+}
```
Panquesito7, line 61:
```suggestion
}

```

bit_manipulation/count_of_set_bits2.cpp
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * @brief Main function
+ * @returns 0 on exit
+ */
+int main() {
+    test();  // run self-test implementations
+    return 0;
+}
```
Panquesito7, line 69: Missing an endline. Without this endline, there could be issues with some IDEs and configurations.
```suggestion
}

```

Panquesito7 (changes requested):
Nice work! 🚀
Please enable [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions) in your repository of this fork in this link: https://github.com/deoliaman00/C-Plus-Plus/actions

bit_manipulation/count_of_set_bits2.cpp
deoliaman00: Did you meant this 
int main() {
    test();  // run self-test implementations
    std::cout << std::endl;
    return 0; 
}

PR #2115, opened 2022-10-05
fix: karatsuba's algorithm is not compiling

### LINKED ISSUE #2086: [BUG] Karatsuba's algorithm is not compiling
### Description

If you try compiling [Karatsuba's algorithm](https://github.com/TheAlgorithms/C-Plus-Plus/tree/master/divide_and_conquer), it will crash with the Segmentation Fault error message.
I've tried various methods and things, but nothing has worked.

It seems the size is overflowing. This seems to be the problem: either the integer types are too small, or something else is wrong.

https://github.com/TheAlgorithms/C-Plus-Plus/blob/24f68ea4dbfda39f36a04864371ce706648383a4/divide_and_conquer/karatsuba_algorithm_for_fast_multiplication.cpp#L121-L126

### Expected behavior

The program should compile and pass the tests as excepted.

### Actual behavior

The program crashes with a Segmentation Fault error message.

### Possible fix

_No response_

### Steps to reproduce

1. Compile the program, either locally or using [Gitpod](https://gitpod.io/#https://github.com/TheAlgorithms/C-Plus-Plus).
2. You will see the program fails with a Segmentation Fault error.
3. There's another minor `clang-tidy` error/warning that can be fixed by replacing all `int64_t` instances with `uint64_t`. That will give another overflow issue; something similar to this.

```bash
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::at: __n (which is 18446744073709551615) >= this->size() (which is 1)
Aborted
```

### Context

This is unprofessional. We should verify all of our implementations work fine.

### Additional information

If you've found a solution to this, feel free to submit a PR! Thank you. 🙂

poypoyan:
doc: improved comments

#### Description of Change
Fixes #2086. There are two problems found:
1. Wrong usage of `std::to_string`. This function outputs the ASCII value of `char` instead of converting char bit to string. A fix (among many other ways) is to replace it with `std::string::insert`.
2. Since string '0' is not equal to '00', '000', ..., base case of algorithm `if (n == 1) {` may not be reached. The fix is to create a function named `safe_substr` that is used to carefully split the bit string with consideration to leading zeros.  

#### Checklist

- [ ] ~Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#New-File-Name-guidelines)~
- [x] ~Added tests and example, test must pass~
- [ ] ~Added documentation so that the program is self-explanatory and educational - [Doxygen guidelines](https://www.doxygen.nl/manual/docblocks.html)~
- [ ] ~Search previous suggestions before making a new one, as yours may be a duplicate.~

Notes:

### DIFF
--- a/divide_and_conquer/karatsuba_algorithm_for_fast_multiplication.cpp
+++ b/divide_and_conquer/karatsuba_algorithm_for_fast_multiplication.cpp
@@ -25,7 +25,7 @@ namespace divide_and_conquer {
 /**
  * @namespace karatsuba_algorithm
  * @brief Functions for the [Karatsuba algorithm for fast
- * multiplication](https://en.wikipedia.org/wiki/Karatsuba_algorithm)
+ * multiplication](https://en.wikipedia.org/wiki/Karatsuba_algorithm) implementation
  */
 namespace karatsuba_algorithm {
 /**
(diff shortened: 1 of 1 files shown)


Panquesito7 (approved):
Awesome! Thank you. 🚀

mishraabhinn (approved):

PR #2120, opened 2022-10-05
fix formulas and documentation of permimeter.cpp

eshaanagarwal:
#### Description of Change

- [x] updated and corrcted documentation of perimeter of traingle and paralleogram
traingle needed documentation for hypotenuse

#### Checklist

Notes:

### DIFF
--- a/math/perimeter.cpp
+++ b/math/perimeter.cpp
@@ -45,12 +45,13 @@ T rect_perimeter(T length, T width) {
  * @brief perimeter of a [triangle](https://en.wikipedia.org/wiki/Triangle) (a +
  * b + c)
  * @param base is the length of the bottom side of the triangle
- * @param height is the length of the tallest point in the triangle
+ * @param left_side is the length of the left side corresponding to the base of triangle
+ * @param right_side is the length of the right side corresponding to the base of triangle
  * @returns perimeter of the triangle
  */
 template <typename T>
-T triangle_perimeter(T base, T height, T hypotenuse) {
-    return base + height + hypotenuse;
+T triangle_perimeter(T base, T left_side, T right_side) {
+    return base + left_side + right_side;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


Panquesito7 (approved):
Looks good. Thanks! 🎉

math/perimeter.cpp
```diff
@@ -45,12 +45,13 @@ T rect_perimeter(T length, T width) {
...
  * @param base is the length of the bottom side of the triangle
- * @param height is the length of the tallest point in the triangle
+ * @param left_side is the length of the left side corresponding to the base of triangle
+ * @param right_side is the length of the right side corresponding to the base of triangle
  * @returns perimeter of the triangle
  */
 template <typename T>
-T triangle_perimeter(T base, T height, T hypotenuse) {
-    return base + height + hypotenuse;
+T triangle_perimeter(T base, T left_side, T right_side) {
```
kvedala, line 53: I prefer to see the "height & base" instead of "left and right" and the former is more intuitive.

math/perimeter.cpp
```diff
@@ -102,14 +103,14 @@ T n_polygon_surface_perimeter(T sides, T length) {
 
 /**
  * @brief surface perimeter of a
- * [cylinder](https://en.wikipedia.org/wiki/Cylinder) (2 * radius + 2 * height)
+ * [cylinder](https://en.wikipedia.org/wiki/Cylinder) 2 * (2 * pi * radius + height)
```
kvedala, line 106:
```suggestion
 * [cylinder](https://en.wikipedia.org/wiki/Cylinder)  \f$2\times\textrm{radius} + 2 \times\textrm{height}\f$
```
please note that the formula implemented in the function should be the same that in the comments.

kvedala (commented):
Based on the [article](https://en.wikipedia.org/wiki/Right_triangle), the two sides of a right triangle forming the right angle are called the **legs** of the triangle. Personally, I find the terms *base* and *height* to be more intuitive than *left side* and *right side* - can triangle have a left side and a right side? 🤔

eshaanagarwal:
> Based on the [article](https://en.wikipedia.org/wiki/Right_triangle), the two sides of a right triangle forming the right angle are called the **legs** of the triangle. Personally, I find the terms _base_ and _height_ to be more intuitive than _left side_ and _right side_ - can triangle have a left side and a right side? 🤔

I have used left and right side with reference to a base. If we can say that base is the bottom side then we can actually determine left and right with reference to it. 
But I guess we should go ahead with base and height then. I will make those changes

math/perimeter.cpp
eshaanagarwal: Oops missed that somehow. Will make that change

math/perimeter.cpp
krishnacx: ` return base + height + hypotenuse;` This was apt

PR #2131, opened 2022-10-05
perfect_number.cpp

Izhar03:
#### Description of Change

I have added a new file/code on checking if a number is perfect or not

#### Checklist

Notes: 
My perfect number does not contain any test cases as such so please have a look on that


math/perfect_number.cpp
```diff
@@ -0,0 +1,58 @@
+#include <iostream>
+
+using namespace std;
+
+void Check_Perfect_Number(int number, int i, int sum) {
```
martin-sos, line 5: to be usable as a library function, Check_Perfect_Number shall not print the result but rather indicating it via a suitable return value. A boolean seems fine, I think. 

The function also depends on the paramter i and sum although this is not really needed. To keep cohesion low we could remove these parameters.

math/perfect_number.cpp
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  if (sum == number) // 
+    cout << i << " is a perfect number\n";
+  else
+    cout << i << " is not a perfect number\n";
+
+}
+
+int main() {
```
martin-sos, line 20: You can automatically test your implementation with the first 7 perfect numbers: https://en.wikipedia.org/wiki/List_of_Mersenne_primes_and_perfect_numbers

math/perfect_number.cpp
```diff
@@ -0,0 +1,58 @@
+#include <iostream>
+
+using namespace std;
+
+void Check_Perfect_Number(int number, int i, int sum) {
+
+  while (i < number) { // to check if the number itself is not less than 1
```
martin-sos, line 7: "i" does not need to be incremented until number, it is enough to run until i < number / 2 +1, as number above that are no dividers anyway

PR #2224, opened 2022-10-12
feat: add power of two algorithm

PraffulGupta-25811:
Added new CPU-Scheduling Algorithm Shortest Job First


Panquesito7 (dismissed):
Awesome. Thanks! 🚀

tjgurwara99 (approved):

bit_manipulation/power_of_2.cpp
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+#include <cassert>   /// for assert
+#include <iostream>  /// for IO operations
+
+/**
+ * @namespace bit_manipulation
+ * @brief Bit manipulation algorithms
+ */
+namespace bit_manipulation {
```
krishnacx, line 24: Why 2 namespaces are required ? Is it like you would add more algorithms in this file ?

krishnacx (dismissed):

bit_manipulation/power_of_2.cpp
Panquesito7: Actually, only two namespaces are required in case there are a lot of functions. But in this case, it's only one function, so we can use `bit_manipulation` only.

Panquesito7 (approved):

PR #2228, opened 2022-10-12
Create Longest_common_prefix.cpp

NEELAMSINGHH:
#### Checklist

- [ ] Added description of change

Notes:

### DIFF
--- a/strings/Longest_common_prefix.cpp
+++ b/strings/Longest_common_prefix.cpp
@@ -1,51 +1,47 @@
-//  A C++ Program to find the longest common prefix
-#include<bits/stdc++.h>
+// C++ program to find longest common prefix
+// of given array of words.
+#include<iostream>
+#include<algorithm>
+
 using namespace std;
 
-// A Utility Function to find the common prefix between
-// strings- str1 and str2
-string commonPrefixUtil(string& str1, string& str2)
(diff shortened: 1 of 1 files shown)


strings/Longest_common_prefix.cpp
```diff
@@ -0,0 +1,51 @@
...
+        if (str1[i] != str2[i])
+            break;
+        result += str1[i];
+    }
+
+    return (result);
+}
+
+// A Function that returns the longest common prefix
+// from the array of strings
```
martin-sos, line 24: This function does not detect the _longest_ common prefix. 
The output of this program is
> The longest common prefix is - gee
Accorfing to the desciption I would expect it to be 
> The longest common prefix is - geeks

It would make sense qualify the array as const, as writing into it is not necessary.

strings/Longest_common_prefix.cpp
```diff
@@ -0,0 +1,51 @@
+//  A C++ Program to find the longest common prefix
+#include<bits/stdc++.h>
+using namespace std;
+
```
martin-sos, line 4: As the functions do only read strings it would make sense to qualify the string arguments as const.

NEELAMSINGHH: i'm updating for the same

PR #2278, opened 2022-10-16
c-plus-plus update kadane algo

MONU8OO8:
I have update the kadane algorithm. there is a error that show max is not define so for that I make max function which I use in kadane algorithm .
thanks for reading my pull request!


Panquesito7:
Sorry, but this does not fit our standards. Thank you for contributing. 🙂

PR #2338, opened 2022-10-21
Added an if statemnt in the mergerSort function

Subhajit098:
Added an if statement in the mergeSort function to whether the first and last index of the array are same or not.


Subhajit098:
Sir please approve it.

PR #2361, opened 2022-10-25
Create policemen-catch-thieves

Aayushow1912:
#### Description of Change
Given an array of size n that has the following specifications: 

Each element in the array contains either a policeman or a thief.
Each policeman can catch only one thief.
A policeman cannot catch a thief who is more than K units away from the policeman.
We need to find the maximum number of thieves that can be caught.


SpiderMath (changes requested):
Your pull request does not follow the file naming guidelines. Please take a look at them [here](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#new-file-name-guidelines)

PR #2404, opened 2022-12-05
Minor changes to improve readability in graph/depth_first_search.cpp

tastybelgianfries:
#### Description of Change

1. In some places, one argument was passed as reference while another was passed as a pointer. I changed such arguments that used pointers  to references to maintain uniformity in code and enhance readability. Related code is changed to work with references.
2. Removed a grammatical/spelling error in a comment on line 48.

#### Checklist

### DIFF
--- a/graph/depth_first_search.cpp
+++ b/graph/depth_first_search.cpp
@@ -45,7 +45,7 @@
 namespace graph {
 /**
  * \brief
- * Adds and edge between two vertices of graph say u and v in this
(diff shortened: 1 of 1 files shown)


Panquesito7 (changes requested):
I think we should stick to pointers because:

1. It prevents a CI warning.
2. With pointers, it's the same as passing it as a reference.

Would you like to add self-tests instead? Thanks. 🙂

PR #2408, opened 2022-12-16
feat: improve the Awesome Workflow

Panquesito7:
#### Description of Change

- Fix Awesome Workflow issues.
  - Thanks to @tjgurwara99 for the original fix: TheAlgorithms/C#1176

#### Checklist

- [ ] ~Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#New-File-Name-guidelines)~
- [ ] ~Added tests and example, test must pass~
- [ ] ~Added documentation so that the program is self-explanatory and educational - [Doxygen guidelines](https://www.doxygen.nl/manual/docblocks.html)~
- [ ] ~Relevant documentation/comments is changed or added~

Notes:

### DIFF
--- a/.github/workflows/awesome_workflow.yml
+++ b/.github/workflows/awesome_workflow.yml
@@ -15,8 +15,8 @@ jobs:
       - uses: actions/setup-python@v2
       - name: requirements
         run: | 
-          sudo apt -qq -y update
-          sudo apt -qq install clang-tidy-10 clang-format-10
+          sudo apt-get -qq update
+          sudo apt-get -qq install clang-tidy clang-format
         # checks are passing with less errors when used with this version. 
         # The default installs v6.0 which did not work out well in my tests
       - name: Setup Git Specs
--- a/.github/workflows/awesome_workflow.yml
+++ b/.github/workflows/awesome_workflow.yml
@@ -8,8 +8,10 @@ jobs:
     name: Code Formatter
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1 # v2 is broken for git diff
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v4
       - name: requirements
         run: | 
           sudo apt-get -qq update
@@ -20,29 +22,11 @@ jobs:
         run: |
           git config --global user.name github-actions[bot]
           git config --global user.email 'github-actions@users.noreply.github.com'
-          git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY
       - name: Filename Formatter
         run: |
-          IFS=$'\n'
-          for fname in `find . -type f -name '*.cpp' -o -name '*.cc' -o -name '*.h'`
-          do 
-            echo "${fname}"
-            new_fname=`echo ${fname} | tr ' ' '_'`
-            echo "      ${new_fname}"
-            new_fname=`echo ${new_fname} | tr 'A-Z' 'a-z'`
-            echo "      ${new_fname}"
-            new_fname=`echo ${new_fname} | tr '-' '_'`
-            echo "      ${new_fname}"
-            new_fname=${new_fname/.cc/.cpp}
-            echo "      ${new_fname}"
-            if [ ${fname} != ${new_fname} ]
-            then
-              echo "      ${fname} --> ${new_fname}"
-              git "mv" "${fname}" ${new_fname}
-            fi
-          done
-          git commit -am "formatting filenames ${GITHUB_SHA::8}" ||  true
-
+          wget https://raw.githubusercontent.com/TheAlgorithms/scripts/main/filename_formatter.sh
+          chmod u=rwx,g=r,o=r filename_formatter.sh
+          ./filename_formatter.sh . .cpp,.hpp
       - name: Update DIRECTORY.md
         run: |
           wget https://raw.githubusercontent.com/TheAlgorithms/scripts/main/build_directory_md.py
@@ -50,9 +34,7 @@ jobs:
           git commit -m "updating DIRECTORY.md" DIRECTORY.md ||  true
       - name: Get file changes
         run: |
-          git remote -v
           git branch
-          git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY
           git diff --diff-filter=dr --name-only origin/master > git_diff.txt
           echo "Files changed-- `cat git_diff.txt`"
       - name: Configure for static lint checks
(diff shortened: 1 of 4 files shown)


.github/workflows/awesome_workflow.yml
```diff
@@ -15,8 +15,8 @@ jobs:
       - uses: actions/setup-python@v2
       - name: requirements
         run: | 
-          sudo apt -qq -y update
-          sudo apt -qq install clang-tidy-10 clang-format-10
+          sudo apt-get -qq update
+          sudo apt-get -qq install clang-tidy clang-format
```
tjgurwara99, line 19: For some reason I can't add suggestion to the whole file, but this looks good except the use of your personal credentials. I would rather you use the `permissions` block at the top level scope (where `jobs` is defined).

For example:

```yml
name: Awesome CI Workflow
on: [push, pull_request]
permissions:
  contents: write # this should suffice I think
jobs:
  MainSequence:
    name: Code Formatter
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v1 # v2 is broken for git diff
       - uses: actions/setup-python@v2
       - name: requirements
         run: | 
           sudo apt-get -qq update
           sudo apt-get -qq install clang-tidy clang-format
        # checks are passing with less errors when used with this version. 
         # The default installs v6.0 which did not work out well in my tests
       - name: Setup Git Specs
        run: |
          git config --global user.name github-actions[bot] # this and the email should be different from an actual user
          git config --global user.email 'github-actions@users.noreply.github.com'
          git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY
     ...
```

.github/workflows/codeql_analysis.yml
```diff
@@ -1,48 +1,56 @@
...
     - name: Checkout repository
-      uses: actions/checkout@main
-
-    # Initializes the CodeQL tools for scanning.
+      uses: actions/checkout@v3
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@main
+      uses: github/codeql-action/init@v2
       with:
         languages: ${{ matrix.language }}
```
tjgurwara99, line 28: This is a matrix build on a single language - just use `cpp` here maybe?

```suggestion
        languages: cpp
```

.github/workflows/codeql_analysis.yml
```diff
@@ -1,48 +1,56 @@
...
+    #   echo "Run, Build Application using script"
+    #   ./location_of_script_within_repo/buildscript.sh
+    #
+    #   In our case, this would be a CMake build step.
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@main
+      uses: github/codeql-action/analyze@v2
+      with:
+        category: "/language:${{matrix.language}}"
```
tjgurwara99, line 56:
```suggestion
        category: "/language:cpp"
```

tjgurwara99 (commented):
LGTM just mentioned a few things - let me know if you don't care about those, can approve it 😄

.github/workflows/awesome_workflow.yml
Panquesito7: I didn't know that was possible 🤔 this is the reason I was using my credentials to make it easier 😅 A few weeks back I read about this and said this wasn't possible. Let's give it a try, though.

.github/workflows/awesome_workflow.yml
```diff
@@ -15,14 +12,14 @@ jobs:
...
         # checks are passing with less errors when used with this version. 
         # The default installs v6.0 which did not work out well in my tests
       - name: Setup Git Specs
         run: |
-          git config --global user.name David Leal
-          git config --global user.email 'Panquesito7@users.noreply.github.com'
+          git config --global user.name github-actions[bot]
+          git config --global user.email 'github-actions@users.noreply.github.com'
           git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY
```
tjgurwara99, line 23: Also this line is not needed now I think - you can just use the default `git push` and it should work.
```suggestion
```

This might also be the reason for all then issues of permissions throughout the workflow - remove these instances (line 58) and it should be fine

tjgurwara99:
Also a general observation that I've made is that this workflow is quite big in itself, I think we can break this down into smaller actions (maybe even a repository for these actions - but that's something bigger). Once this gets merged, I think I can make these smaller and write them in a contained way if you'd like 😄

.github/workflows/awesome_workflow.yml
tjgurwara99: Also, the line 56

```sh
git diff --diff-filter=dr --name-only origin/master > git_diff.txt
```

is quite old way of referencing origin branches and potentially problematic. You should change the `checkout` action to `v3` since then use
```sh
git diff --diff-filter=dr --name-only origin master > git_diff.txt
```

the `/` character in the middle of `origin` and `master` is not correct in newer versions of `git`

.github/workflows/awesome_workflow.yml
```diff
@@ -1,51 +1,30 @@
...
+  contents: write
 
 jobs:
   MainSequence:
     name: Code Formatter
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1 # v2 is broken for git diff
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
```
tjgurwara99, line 11: You need the `fetch-depth` here otherwise it would fail...

```suggestion
      - uses: actions/checkout@v3
        with:
          fetch-depth: 0
```

tjgurwara99 (changes requested):

.github/workflows/awesome_workflow.yml
```diff
@@ -1,51 +1,32 @@
...
       - name: Setup Git Specs
         run: |
-          git config --global user.name David Leal
-          git config --global user.email 'Panquesito7@users.noreply.github.com'
-          git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY
+          git config --global user.name github-actions[bot]
+          git config --global user.email 'github-actions@users.noreply.github.com'
+
+          git checkout HEAD:$GITHUB_REF
+          git pull || true # Always fetch changes before making any changes
```
tjgurwara99, line 25: You probably don't need that with `fetch-depth: 0` enabled.

Panquesito7:
Seems to be working now. Thanks a ton for your help. 🙂
Will change the workflow to do the same thing in the C repository.

.github/workflows/awesome_workflow.yml
```diff
@@ -1,61 +1,40 @@
...
-            if [ ${fname} != ${new_fname} ]
-            then
-              echo "      ${fname} --> ${new_fname}"
-              git "mv" "${fname}" ${new_fname}
-            fi
-          done
-          git commit -am "formatting filenames ${GITHUB_SHA::8}" ||  true
-
+          wget https://raw.githubusercontent.com/TheAlgorithms/scripts/main/filename_formatter.sh
+          chmod u=rwx,g=r,o=r filename_formatter.sh
```
tjgurwara99, line 28:
```suggestion
          chmod +x filename_formatter.sh
```

tjgurwara99 (approved):
LGTM 😄 We're at a good place now and can iterate 😄

Panquesito7:
> I'll open a few PR's in the `scripts` repository (hopefully this weekend) so that we can make these scripts more general

Cool! We can discuss this via Discord. Thanks. 🙂

Panquesito7:
https://github.com/TheAlgorithms/C/actions/runs/3728181169/jobs/6322964878#step:11:19
Still seems to be failing for that branch protection check. 🤔

tjgurwara99:
I guess that's why you needed you handle to merge it. This is a problem, I think I might have a solution but I will write up what I mean after work. In the meantime, we should open an issue for this instead 😄

Issue #2456, opened 2023-04-30
[FIX] Pre-existing algorithms do not fit the contribution guidelines


realstealthninja:
### Detailed description

<h1 align="center">Times change </h1>  
<p align="center">so does the development of algorithms and documentation</p>

Some of the older algorithms existing in this repository do not fit the [contribution guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md).

A short list of examples:
 - [Armstrong number](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/math/armstrong_number.cpp)
 - [binary exponent](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/math/binary_exponent.cpp)
 - [check prime](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/math/check_prime.cpp) 
 - [complex numbers](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/math/complex_numbers.cpp)
 - [square root double](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/math/sqrt_double.cpp)

### Context

Educating new developers on coding standards is a major priority; As the industry today relies heavily on documentation;
If you were to ask any API developer/user on what the most important part of their development and use it will always be:
 - Code quality
 - documentation

having a fast and performant algorithm is good but whats the point if you cant understand how it works or even why it works.

I think it is in the best interest of us to document our algorithms.

As i consider all algorithms must be:
- beautiful
- understandable
- educational

### Additional information

Others in the past have made major changes like this to this repository and i think its time for us to do the same. to keep the torch moving.

An homage to [@kvedala] for their part in trying to standardize this repository #805

Kunal766:
Hi,  I wanted to start contributing to this repository ,can you assign me this problem

realstealthninja:
No sorry no assigning in this repository just start making a pull request and we will review to help you.❤️

Note: assigning requires triage access

CodeSinghh:
how can help so that this issue will be resolved

realstealthninja:
> how can help so that this issue will be resolved
just find an algorithm that does not fit the contribution guidelines and start fixing it! we will close this issue once *ALL* the algorithms are fixed! :smile:

ewd00010:
# Math Directory Checklist
### (pending update)
Below is the list of algorithms to be checked within the maths repository.
- [X] aliquot_sum.cpp
[...]

PR #2463, opened 2023-05-16
fix:fit `binary-exponent.cpp` to guidelines

ewd00010:
#### Description of Change
FIX: added checking for negative exponents to both functions. Removed user interaction and replaced with tests function.

CHORE: added author, reformatted function documentation, added descriptions of what is happening in each algo and adhered to other repo standards

#### Checklist

Notes:  Should we keep formatting text like `\f$O(\log(n))\f$` in the comments or should they be removed?
Additionally i cant seem to get clang-tidy to work on my system, this code will need a once over with it


math/binary_exponent.cpp
```diff
@@ -19,29 +20,71 @@
  * \f}
  * Hence to calculate 2^10 we only need to multiply \f$2^8\f$ and \f$2^2\f$
  * skipping \f$2^1\f$ and \f$2^4\f$.
+ *
+ * @author[Mann Mehta]https://github.com/mann2108
  */
 
-#include <iostream>
+#include <cassert> // for assert
+#include <iostream> // for cout
```
realstealthninja, line 29:
```suggestion
/**
 * @brief Mathematical algorithms
 * @namespace
 */
```

realstealthninja (changes requested):

math/binary_exponent.cpp
```diff
@@ -19,29 +20,71 @@
...
```
realstealthninja, line 28:
```suggestion
#include <cassert> /// for assert
#include <iostream> /// for cout
```

realstealthninja (changes requested):

realstealthninja:
We use \f$ to render latex math equations they are necessary and imo should be required as they make the equations professional and presentable for more details check out doxygen's documentation on  latex equations

realstealthninja:
part of #2456

realstealthninja (approved):
Looks good! thank you so much for your contributions :heart:

math/binary_exponent.cpp
```diff
@@ -19,29 +20,74 @@
  * \f}
  * Hence to calculate 2^10 we only need to multiply \f$2^8\f$ and \f$2^2\f$
  * skipping \f$2^1\f$ and \f$2^4\f$.
+ *
+ * @author[Mann Mehta]https://github.com/mann2108
```
Panquesito7, line 24: Wrong Markdown link usage.
```suggestion
 * @author [Mann Mehta](https://github.com/mann2108)
```

math/binary_exponent.cpp
```diff
@@ -19,29 +20,74 @@
...
  */
 
-#include <iostream>
-
-/// Recursive function to calculate exponent in \f$O(\log(n))\f$ using binary
-/// exponent.
-int binExpo(int a, int b) {
+#include <cassert> /// for assert
+#include <iostream> /// for cout
```
Panquesito7, line 28:
```suggestion
#include <cassert>  /// for assert
#include <iostream> /// for IO operations

```

math/binary_exponent.cpp
```diff
@@ -19,29 +20,74 @@
...
+ * binary exponent.
+ * @param a base number
+ * @param b exponent number
+ * @return 1 if the exponent number is 0 the return value will always be 1
+ * @return 0 if the exponential number is < 0 ( 0 representing NULL since
+ * < 0 is an invalid exponential to use )
+ * @return (res * res * a) if modulo 2 of exponential is 0
+ * @return (res * res) if modulo 2 of exponential is 1
+ */
+long long binExpo_recursive(long long a, long long b) {
```
Panquesito7, line 45: Consider using `uint64_t` for non-negative values (or their appropriate size: `uint32_t`, `uint16_t`, `uint8_t`) or `int64_t` for negative values. Check other parts of the code ([reference](https://en.cppreference.com/w/cpp/types/integer)). **Let us know if you need any help here or on our [Discord server](https://the-algorithms.com/discord).** 🙂

math/binary_exponent.cpp
```diff
@@ -19,29 +20,74 @@
...
 }
 
-/// Iterative function to calculate exponent in \f$O(\log(n))\f$ using binary
-/// exponent.
-int binExpo_alt(int a, int b) {
-    int res = 1;
-    while (b > 0) {
+/**
+ * @brief Iterative function to calculate exponent in \f$O(\log(n))\f$ using
+ binary
```
Panquesito7, line 71:
```suggestion
 * binary
```

math/binary_exponent.cpp
```diff
@@ -50,22 +96,39 @@ int binExpo_alt(int a, int b) {
...
-    /// Give two numbers a, b
-    std::cin >> a >> b;
-    if (a == 0 && b == 0) {
-        std::cout << "Math error" << std::endl;
-    } else if (b < 0) {
-        std::cout << "Exponent must be positive !!" << std::endl;
-    } else {
-        int resRecurse = binExpo(a, b);
-        /// int resIterate = binExpo_alt(a, b);
+using namespace math;
```
Panquesito7, line 101: It doesn't harm if you use `math` directly. This looks unclean and unprofessional, IMO.

Panquesito7 (changes requested):
For more details, you can check the [`clang-tidy`](https://github.com/TheAlgorithms/C-Plus-Plus/actions/runs/5059924117/jobs/9225076950?pr=2463) warnings.
If you need any help fixing those, let us know. 🙂

math/binary_exponent.cpp
ewd00010: can you expand on unprofessional? I'm assuming unclean is because using's are meant to be at the top of the file ( can't remember why I put it on this line ). I can't seem to see another issue; the intention I had was to not use a lot of scope resolution operators, making it a little easier to read.

math/binary_exponent.cpp
```diff
@@ -19,29 +20,75 @@
...
+ * binary exponent.
+ * @param a base number
+ * @param b exponent number
+ * @return 1 if the exponent number is 0 the return value will always be 1
+ * @return 0 if the exponential number is < 0 ( 0 representing NULL since
+ * < 0 is an invalid exponential to use )
+ * @return (res * res * a) if modulo 2 of exponential is 0
+ * @return (res * res) if modulo 2 of exponential is 1
+ */
+uint64_t binExpo_recursive(int64_t a, int64_t b) {
```
tjgurwara99, line 46: I would say to keep the change in type casting to a minimum - since you are changing the return type to use the `uint64_t` type then make your arguments `uint64_t` type too. The added benefit is that you don't need to add another check in your code about the valid exponent `if (b < 0)` and it would be less prone to type casting issues.

A general advice for C/C++ (it doesn't really do much to this small scale program but should still be relevant) - type casting should be avoided until you really need it and you know what you're doing...

math/binary_exponent.cpp
```diff
@@ -50,22 +97,38 @@ int binExpo_alt(int a, int b) {
...
-    }
+    assert(math::binExpo_iterative(1, 0) == 1);
+    assert(math::binExpo_iterative(746584, 0) == 1);
+    assert(math::binExpo_iterative(1, 1) == 1);
+    assert(math::binExpo_iterative(2938374, 1) == 2938374);
+    assert(math::binExpo_iterative(3, 7) == 2187);
+    assert(math::binExpo_iterative(31, 5) == 28629151);
+    assert(math::binExpo_iterative(0, 0) == 1);
+    assert(math::binExpo_iterative(1, -20) == 0);
+    std::cout << "All tests have successfully passed!" << std::endl;
```
tjgurwara99, line 124: Can you consider adding a benchmark to showcase the differences between these two approaches when it comes to speed and memory consumption? Not necessary for me to approve this PR but its a good thing for people learning to use the language.

math/binary_exponent.cpp
tjgurwara99: Scope resolution is a good thing for a few reasons and I don't think I need to go into the details here, but I would just give you two reasons why `using namespace abc` is a bad idea:

1. Name collisions - the main reason namespacing was created, consider that you are using the classic `using namespace std;` and you created a `string` class in your current `namespace abc`, `using namespace std;` would cause a naming collision which actually is very difficult to resolve when you have a big dependency chain (in small scale programs you won't notice any problems but once you get to bigger projects you would start to dislike people who use it in their code 😂 jkjk)
2. Debugging problems - this also applies to bigger codebases, the `using namespace abc` would make your life very difficult when using `gdb` or any other debugging tool (that is not paid for).

For the matter of "unprofessional", I think it shows that the coder is novice so that's a different thing from unprofessional but still should be relevant 😄

math/binary_exponent.cpp
```diff
@@ -19,53 +20,116 @@
  * \f}
  * Hence to calculate 2^10 we only need to multiply \f$2^8\f$ and \f$2^2\f$
  * skipping \f$2^1\f$ and \f$2^4\f$.
+ *
+ * @author [Mann Mehta](https://github.com/mann2108)
```
Panquesito7, line 24: You can also add yourself as you've helped update this algorithm to our standards. 🙂
```suggestion
 * @author [Mann Mehta](https://github.com/mann2108)
 * @author [ewd00010](https://github.com/ewd00010)
```

math/binary_exponent.cpp
ewd00010: Tried a couple of ways to get a memory usage check for each function, but no success so far.

Considering using getRUsage, but would require download of Cygwin for a ```sys/resource.h```, not sure if that's worthwhile due to users having to download it just to see a simple memory check. 

For now, feel free to approve the PR and maybe make an issue seeing if someone else can do a better job of this then me. If its still up as an issue after a while, I might try again!

Panquesito7 (approved):
Awesome. Thanks! 🚀

alexpantyukhin (approved):

realstealthninja (approved):
LGTM!

realstealthninja:
just one conversation to resolve and you are good to go!

math/binary_exponent.cpp
tjgurwara99: The proper way to run benchmarks is to run the same function with same inputs over a large number of times rather than doing one call and comparing it to the other call, most computers would run the functions with negligible difference and you will get nothing but if you loop it over a large enough N, then you will have meaningful results.

If you don't think it's possible then feel free to mark this as resolved and then I can look at adding some prior art to doing benchmarks to help people learn😄

ewd00010: Sorry I didn't see this until now, I will try to implement it that way at some point and see what the results are. Feel free to do it before me! Can't guarantee a deadline for it.

Panquesito7:
Will be merging in a few days if there are no objections. Ping me in case I forget.

realstealthninja:
@Panquesito7

PR #2512, opened 2023-07-24
Letter Combinations of a Phone Number

sOR-o:
#### Checklist

Added a question on backtracking. 

### DIFF
--- a/backtracking/letter_combinations_of_a_phone_number.cpp
+++ b/backtracking/letter_combinations_of_a_phone_number.cpp
@@ -140,7 +140,6 @@ static void test() {
  * @returns 0 on exit.
  */
 int main() {
-    // Run the test function
-    test();
+    test(); //  run self-test implementations 
     return 0;
 }
--- a/backtracking/letter_combinations_of_a_phone_number.cpp
+++ b/backtracking/letter_combinations_of_a_phone_number.cpp
@@ -25,39 +25,36 @@
 #include <iostream>  /// for IO operations
 #include <vector>    /// for std::vector
 
-/**
- * @namespace backtracking
- * @brief Backtracking algorithms
- */
 namespace backtracking {
-/**
- * @namespace letter_combinations
- * @brief Functions for generating letter combinations from digits using
- * backtracking.
- */
 namespace letter_combinations {
+
+// Helper function to convert a character digit to an unsigned integer
+unsigned toUnsigned(const char digit) {
+    return static_cast<unsigned>(digit - '0');
+}
+
 /**
  * @brief Recursive utility function to generate letter combinations.
  * @param digit The input digit string to generate combinations for.
  * @param output The current output combination being formed.
- * @param index The index of the current digit being processed.
  * @param ans Vector to store the generated combinations.
  * @param mpp Array containing the mapping of digits to letters.
  */
-void solve(const std::string& digit, std::string& output, std::size_t index,
+void solve(const std::string& digit, std::string& output,
            std::vector<std::string>& ans,
            const std::array<std::string, 10>& mpp) {
-    if (index >= digit.length()) {
+    if (output.size() == digit.length()) {
         ans.push_back(output);
         return;
     }
 
-    int num = digit[index] - '0';
-    std::string value = mpp[num];
+    unsigned num = toUnsigned(digit[output.size()]);
+    const std::string& value = mpp[num];
 
-    for (int i = 0; i < value.size(); i++) {
-        output.push_back(value[i]);
-        solve(digit, output, index + 1, ans, mpp);
+    for (unsigned i = 0; i < value.size(); ++i) {
+        char letter = value[i];
+        output.push_back(letter);
+        solve(digit, output, ans, mpp);
         output.pop_back();
     }
 }
@@ -74,12 +71,10 @@ std::vector<std::string> letterCombinations(const std::string& digit) {
         return ans;
     }
 
-    std::string output = "";
-    int index = 0;
-
-    std::array<std::string, 10> mpp = {"",    "",    "abc",  "def", "ghi",
-                                       "jkl", "mno", "pqrs", "tuv", "wxyz"};
-    solve(digit, output, index, ans, mpp);
+    std::string output;
+    const std::array<std::string, 10> mpp = {
+        "", "", "abc", "def", "ghi", "jkl", "mno", "pqrs", "tuv", "wxyz"};
+    solve(digit, output, ans, mpp);
     return ans;
 }
 }  // namespace letter_combinations
@@ -134,7 +129,8 @@ static void test() {
     std::string digits4 = "";
     std::vector<std::string> combinations4 =
         backtracking::letter_combinations::letterCombinations(digits4);
-    assert(combinations4.empty());
+    std::vector<std::string> expected4;  // Empty vector
+    assert(combinations4 == expected4);
 
     std::cout << "All test cases passed!\n";
 }
--- a/backtracking/letter_combinations_of_a_phone_number.cpp
+++ b/backtracking/letter_combinations_of_a_phone_number.cpp
@@ -1,0 +1,150 @@
+/**
+ * @file
+ * @brief Generates letter combinations for given digits using backtracking.
+ *
+ * @details
+ * This code generates all possible letter combinations corresponding to a given
+ * string of digits. It uses the concept of backtracking to explore all possible
+ * combinations of letters for each digit.
+ *
+ * The function letterCombinations takes a string of digits as input and returns
+ * a vector containing all the generated letter combinations. Each digit is
+ * mapped to a set of letters, and the function backtracks to explore all
+ * possible combinations by appending one letter at a time for each digit.
+ *
+ * The main test function demonstrates the correctness of letterCombinations by
+ * running multiple test cases and checking the output against the expected
+ * results. The expected vectors are now initialized one by one for better
+ * clarity.
+ *
+ * @author [Saurabh Kushwaha](https://github.com/sOR-o)
+ */
+
+#include <array>     /// for std::array
+#include <cassert>   /// for assert
+#include <iostream>  /// for IO operations
+#include <vector>    /// for std::vector
+
+/**
+ * @namespace backtracking
+ * @brief Backtracking algorithms
+ */
+namespace backtracking {
+/**
+ * @namespace letter_combinations
+ * @brief Functions for generating letter combinations from digits using
+ * backtracking.
+ */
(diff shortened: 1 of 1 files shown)


madhavgaba (approved):

backtracking/letter_combinations_of_a_phone_number.cpp
```diff
@@ -0,0 +1,150 @@
...
+namespace letter_combinations {
+/**
+ * @brief Recursive utility function to generate letter combinations.
+ * @param digit The input digit string to generate combinations for.
+ * @param output The current output combination being formed.
+ * @param index The index of the current digit being processed.
+ * @param ans Vector to store the generated combinations.
+ * @param mpp Array containing the mapping of digits to letters.
+ */
+void solve(const std::string& digit, std::string& output, int index,
```
vil02, line 47:
```suggestion
void solve(const std::string& digit, std::string& output, std::size_t index,
```
This will help for `-Werror=sign-compare`.

backtracking/letter_combinations_of_a_phone_number.cpp
```diff
@@ -0,0 +1,150 @@
...
+           std::vector<std::string>& ans,
+           const std::array<std::string, 10>& mpp) {
+    if (index >= digit.length()) {
+        ans.push_back(output);
+        return;
+    }
+
+    int num = digit[index] - '0';
```
vil02, line 55: What do you think about extracting the part `digit[index] - '0'` to a separate function, like `unsigned toUnsinged(const char digit)`? The advantage is, that the name suggest what happens here.

backtracking/letter_combinations_of_a_phone_number.cpp
```diff
@@ -0,0 +1,150 @@
...
+    std::string value = mpp[num];
+
+    for (int i = 0; i < value.size(); i++) {
```
vil02, line 58:
```suggestion
    for (std::size_t i = 0; i < value.size(); i++) {
```
As above, helps for `-Werror=sign-compare`.

backtracking/letter_combinations_of_a_phone_number.cpp
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    std::string output = "";
+    int index = 0;
+
+    std::array<std::string, 10> mpp = {"",    "",    "abc",  "def", "ghi",
+                                       "jkl", "mno", "pqrs", "tuv", "wxyz"};
+    solve(digit, output, index, ans, mpp);
+    return ans;
+}
```
vil02, line 84:
```suggestion
std::vector<std::string> letterCombinations(const std::string& digit) {
    std::vector<std::string> ans;

    if (digit.empty()) {
        return ans;
    }

    const std::array<std::string, 10> mpp = {
        "", "", "abc", "def", "ghi", "jkl", "mno", "pqrs", "tuv", "wxyz"};
    std::string output = "";
    solve(digit, output, 0, ans, mpp);
    return ans;
}
```
Formatting was done by clang-format. I suggest to:
- mark `mpp` as `const`,
- remove the `index` variable.

backtracking/letter_combinations_of_a_phone_number.cpp
```diff
@@ -0,0 +1,150 @@
...
+    assert(combinations3 == expected3);
+
+    // Test Case 4 (Empty Input)
+    std::string digits4 = "";
+    std::vector<std::string> combinations4 =
+        backtracking::letter_combinations::letterCombinations(digits4);
+    assert(combinations4.empty());
+
+    std::cout << "All test cases passed!\n";
+}
```
vil02, line 140:
```suggestion
static void test() {
    const std::vector<std::string> expected23 = {"ad", "ae", "af", "bd", "be",
                                                 "bf", "cd", "ce", "cf"};
    assert(backtracking::letter_combinations::letterCombinations("23") ==
           expected23);

    const std::vector<std::string> expected7 = {"p", "q", "r", "s"};
    assert(backtracking::letter_combinations::letterCombinations("7") ==
           expected7);

    const std::vector<std::string> expected9 = {"w", "x", "y", "z"};
    assert(backtracking::letter_combinations::letterCombinations("9") ==
           expected9);

    assert(backtracking::letter_combinations::letterCombinations("").empty());

    std::cout << "All test cases passed!\n";
}
```

vil02 (changes requested):

backtracking/letter_combinations_of_a_phone_number.cpp
sOR-o: yup. done 👍

sOR-o:
all the suggested changes are commited, and a pull request is made.

backtracking/letter_combinations_of_a_phone_number.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include <cassert>   /// for assert
+#include <iostream>  /// for IO operations
+#include <vector>    /// for std::vector
+
+namespace backtracking {
+namespace letter_combinations {
+
+// Helper function to convert a character digit to an unsigned integer
+unsigned toUnsigned(const char digit) {
+    return static_cast<unsigned>(digit - '0');
```
vil02, line 33: The result of this function is used as the input for [`std::string::operator[]`](https://en.cppreference.com/w/cpp/string/basic_string/operator_at), which, in our situation, expects `std::size_t` (sorry, I should have write this earlier).
```suggestion
std::size_t toUnsigned(const char digit) {
    return static_cast<std::size_t>(digit - '0');
```

Consider the situation, when someone sends as `digit` `'a'`. How this function should react? Is that even possible? What if the function `letterCombinations` would be used with an input `"abc"`?

backtracking/letter_combinations_of_a_phone_number.cpp
```diff
@@ -0,0 +1,146 @@
...
+    if (output.size() == digit.length()) {
+        ans.push_back(output);
+        return;
+    }
+
+    unsigned num = toUnsigned(digit[output.size()]);
+    const std::string& value = mpp[num];
+
+    for (unsigned i = 0; i < value.size(); ++i) {
+        char letter = value[i];
```
vil02, line 55:
```suggestion
    const auto letters = mpp[toUnsigned(digit[output.size()])];
    for (const char letter : letters) {
```

backtracking/letter_combinations_of_a_phone_number.cpp
```diff
@@ -0,0 +1,146 @@
...
+        output.pop_back();
+    }
+}
+
+/**
+ * @brief Generates letter combinations for a given digit string.
+ * @param digit The input digit string to generate combinations for.
+ * @returns A vector containing all the generated letter combinations.
+ */
+std::vector<std::string> letterCombinations(const std::string& digit) {
```
vil02, line 67: What do you think about using the name `digits` (not only in this function)?

vil02 (changes requested):
Could you please have a look at [my comment regarding the tests](https://github.com/TheAlgorithms/C-Plus-Plus/pull/2512#discussion_r1276732961)? I would like to know your opinion about thus suggestion.

backtracking/letter_combinations_of_a_phone_number.cpp
sOR-o: Regarding the additional test case with input "1," I believe its inclusion is important to test how the code handles edge cases and single-digit inputs. However, the expected output for this test case is not explicitly provided in the code.
Based on my analysis of the code, the digit "1" is mapped to an empty string, indicating that there are no letters associated with "1." Therefore, I assume the expected output for this test case should be an empty vector, as no letter combinations are possible for the digit "1."

To ensure clarity in the test suite, it would be helpful to either add a comment indicating the expected output for this particular test case or include an explicitly defined vector with empty contents as the expected output.

Overall, I appreciate the attention to edge cases, and I'm here to collaborate and make any necessary improvements to the codebase to enhance its correctness and robustness.
Please let me know your thoughts on the matter, and if you have any specific preferences or expectations regarding the test case in question.

backtracking/letter_combinations_of_a_phone_number.cpp
```diff
@@ -0,0 +1,146 @@
...
+    std::cout << "All test cases passed!\n";
+}
+
+/**
+ * @brief Main function.
+ * @returns 0 on exit.
+ */
+int main() {
+    // Run the test function
+    test();
```
realstealthninja, line 144:
```suggestion
    test(); //  run self-test implementations 
```

realstealthninja (changes requested):

backtracking/letter_combinations_of_a_phone_number.cpp
vil02: The test with `"1"` - great that you have added it.

I would still insist on expressing your tests in a more _compact_ way. Please have a look at your code:
```cpp
// Test Case 3
std::string digits3 = "9";
std::vector<std::string> combinations3 =
    backtracking::letter_combinations::letterCombinations(digits3);
std::vector<std::string> expected3;
expected3.push_back("w");
expected3.push_back("x");
expected3.push_back("y");
expected3.push_back("z");
assert(combinations3 == expected3);
```
There is so much going on here:
- you inform the reader that this is _the test case 3_ (by the way: is this important that this is the `test case 3`? Would it fail if it was `4`? Is there some relationship between `3` and the input of the tested function?),
- you define the variable `digits3`,
- you save the result of your function,
- you define the variable `expected3`,
- you push 4 things to the variable `expected3`,
- you check if `combinations3` is same as `expected3`.

There is nothing bad here: it is a perfect test. But, in my opinion, there are simpler ways to express it.

For example:
```cpp
const std::vector<std::string> expected9 = {"w", "x", "y", "z"};
assert(backtracking::letter_combinations::letterCombinations("9") ==
       expected9);
```

You can even consider to split all of the test cases into separate functions, e.g.
```cpp
void test_with_9() {
    const std::vector<std::string> expected = {"w", "x", "y", "z"};
    assert(backtracking::letter_combinations::letterCombinations("9") ==
           expected);
}
```

By the way, have you heard about _Arrange, Act, Assert_?

realstealthninja (commented):
Please do not mark suggestions as resolved unless youve explained what your thoughts are about them and a maintainer/ other reviewers have explicitly asked you to.

backtracking/letter_combinations_of_a_phone_number.cpp
sOR-o: I appreciate your attention to detail and pointing out the potential error related to the -Werror=sign-compare warning. Ensuring type consistency in comparisons is important for maintaining code correctness and preventing unexpected behavior. Thank you for your input!

PR #2517, opened 2023-07-30
feat: add chinese remainder theorem

HimeshKohad:
#### Checklist

Notes: A C++ program to solve a system of linear congruences using CRT.

### DIFF
--- a/math/chinese_remainder_theorem.cpp
+++ b/math/chinese_remainder_theorem.cpp
@@ -49,7 +49,7 @@ int compute_inverse(int a, int m) {
  * @brief Function to solve a system of linear congruences using the Chinese
  * Remainder Theorem (CRT).
  *
- * @param a Array of integers representing the remainders of the congruences.
+ * @param an Array of integers representing the remainders of the congruences.
  * @param m Array of pairwise coprime integers representing the moduli of the
  * congruences.
  * @param n Number of congruences in the system.
--- a/math/chinese_remainder_theorem.cpp
+++ b/math/chinese_remainder_theorem.cpp
@@ -20,6 +20,7 @@
 #include <cassert>   /// for assert
 #include <iostream>  /// for IO operations
 #include <vector>    /// for std::vector
+#include <stdexcept> // for handling exceptions
 
 struct LinearCongruance {
     const int a;
@@ -32,6 +33,9 @@ int compute_product_of_all_moduli(
     const std::vector<LinearCongruance> &congruances) {
     int res = 1;
     for (const auto congruance : congruances) {
+        if (congruance.m == 0) {
+            throw std::invalid_argument("Modulus 'm' cannot be zero.");
+        }
         res *= congruance.m;
     }
     return res;
@@ -84,7 +88,9 @@ void test() {
         LinearCongruance(2, 3), LinearCongruance(3, 4), LinearCongruance(1, 5)};
     assert(chinese_remainder_theorem(testCase) == 11);
 
-    std::cout << "One test has successfully passed!\n";
+    // Add additional test cases here if needed
+
+    std::cout << "All tests have passed successfully!\n";
 }
 
 /**
@@ -92,6 +98,11 @@ void test() {
  * @returns 0 on exit
  */
 int main() {
-    test();  // run self-test implementations
+    try {
+        test(); // run self-test implementations
+    } catch (const std::exception &ex) {
+        std::cerr << "Error: " << ex.what() << std::endl;
+        return 1; // Return a non-zero value to indicate an error
+    }
     return 0;
 }
--- a/math/chinese_remainder_theorem.cpp
+++ b/math/chinese_remainder_theorem.cpp
@@ -36,16 +36,17 @@ struct LinearCongruance {
      * @param a The remainder of the congruence (x == a mod m)
      * @param m The modulus of the congruence (x == a mod m)
      */
-    LinearCongruance(int a, int m) : a(a), m(m) {}
+    LinearCongruance(int a, int m) : a(a), m(m) {
+        if (m == 0) {
+            throw std::invalid_argument("Modulus 'm' cannot be zero.");
+        }
+    }
 };
 
 int compute_product_of_all_moduli(
     const std::vector<LinearCongruance> &congruances) {
     int res = 1;
     for (const auto congruance : congruances) {
-        if (congruance.m == 0) {
-            throw std::invalid_argument("Modulus 'm' cannot be zero.");
-        }
         res *= congruance.m;
     }
     return res;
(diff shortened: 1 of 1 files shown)


math/chinese_remainder_theorem.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * @brief Main Function
+ * @returns 0 on exit
+ */
+int main() {
+    test();  // run self-test implementations
+    return 0;
+}
```
vil02, line 80: The biggest change in my suggestion is to introduce the `struct` `LinearCongruance` storing the `a` and `m` of each congruence in the system. It allows to simplify the signature of the function `chinese_remainder_theorem` and simplify the logic inside (one does not need to worry about cases, when the input arrays have different sizes etc.; it also creates _tighter_ relation between `a`'s and `m`'s).

Instead of witting comments, I suggest to use functions with _meaningful_ names.

The remaining changes are just `clang-format` and some minor updates.

```suggestion
/**
 * @file
 * @brief C++ program to solve a system of linear congruences using the Chinese
 * Remainder Theorem (CRT).
 *
 * @details
 * The Chinese Remainder Theorem (CRT) is a mathematical theorem that provides a
 * solution to a system of linear congruences. It allows us to find a unique
 * solution to the system if the moduli are pairwise coprime (i.e., the greatest
 * common divisor of any two moduli is 1).
 *
 * This implementation assumes that the moduli are pairwise coprime.
 *
 * @note For the system of congruences to have a solution, the moduli must be
 * pairwise coprime. If they are not, the result obtained may not be valid.
 *
 * @author [Himesh](https://github.com/HimeshKohad)
 */

#include <cassert>   /// for assert
#include <iostream>  /// for IO operations
#include <vector>    /// for std::vector

struct LinearCongruance {
    const int a;
    const int m;

    LinearCongruance(int a, int m) : a(a), m(m) {}
};

int compute_product_of_all_moduli(
    const std::vector<LinearCongruance> &congruances) {
    int res = 1;
    for (const auto congruance : congruances) {
        res *= congruance.m;
    }
    return res;
}

int compute_inverse(int a, int m) {
    int inv = 1;
    while ((a * inv) % m != 1) {
        inv++;
    }
    return inv;
}

/**
 * @brief Function to solve a system of linear congruences using the Chinese
 * Remainder Theorem (CRT).
 *
 * @param a Array of integers representing the remainders of the congruences.
 * @param m Array of pairwise coprime integers representing the moduli of the
 * congruences.
 * @param n Number of congruences in the system.
 * @return The solution to the system of congruences.
 */
int chinese_remainder_theorem(
    const std::vector<LinearCongruance> &congruances) {
    // Calculate the product of all moduli
    const auto M = compute_product_of_all_moduli(congruances);

    int result = 0;
    for (const auto congruance : congruances) {
        const auto factor = M / congruance.m;
        const auto inv = compute_inverse(factor, congruance.m);

        // Sum up the partial results
        result += (congruance.a * factor * inv) % M;
    }

    // Make sure the result is non-negative
    result = (result + M) % M;

    return result;
}

/**
 * @brief Self-test implementations
 * @returns void
 */
void test() {
    const std::vector<LinearCongruance> testCase = {
        LinearCongruance(2, 3), LinearCongruance(3, 4), LinearCongruance(1, 5)};
    assert(chinese_remainder_theorem(testCase) == 11);

    std::cout << "One test has successfully passed!\n";
}

/**
 * @brief Main Function
 * @returns 0 on exit
 */
int main() {
    test();  // run self-test implementations
    return 0;
}
```

vil02 (changes requested):
Please consider optimizing computation of the _inverse_. What do you think about using [Extended Euclidean algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm) for that?

Does your function work for negative inputs?

I would also suggest to add more test cases.

math/chinese_remainder_theorem.cpp
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @param an Array of integers representing the remainders of the congruences.
+ * @param m Array of pairwise coprime integers representing the moduli of the
+ * congruences.
+ * @param n Number of congruences in the system.
+ * @return The solution to the system of congruences.
+ */
+int chinese_remainder_theorem(
+    const std::vector<LinearCongruance> &congruances) {
+    // Calculate the product of all moduli
```
vil02, line 60: This comment is not needed.
```suggestion
```

math/chinese_remainder_theorem.cpp
```diff
@@ -0,0 +1,97 @@
...
+ * @author [Himesh](https://github.com/HimeshKohad)
+ */
+
+#include <cassert>   /// for assert
+#include <iostream>  /// for IO operations
+#include <vector>    /// for std::vector
+
+struct LinearCongruance {
+    const int a;
+    const int m;
```
vil02, line 26: What happens if `a` or `m` are negative? What happens if `m` is zero?

vil02 (changes requested):
@HimeshKohad thanks for agreeing with my suggestions. Here are some more.

math/chinese_remainder_theorem.cpp
HimeshKohad: Thank you for your suggestion.
I think the code would work for the cases when a and m are negative because the modular arithmetic operator allows negative numbers to be represented in modulo. 
As for the case where m is zero, I have updated the code accordingly to handle that case.

HimeshKohad (commented):
Tried resolving a few recommended changes.

math/chinese_remainder_theorem.cpp
```diff
@@ -0,0 +1,97 @@
...
+    return res;
+}
+
+int compute_inverse(int a, int m) {
+    int inv = 1;
+    while ((a * inv) % m != 1) {
+        inv++;
+    }
+    return inv;
+}
```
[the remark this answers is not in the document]
HimeshKohad: Thank you for the suggestion, I have implemented the extended euclidean algorithm.
Also, earlier I had assumed that we would always have non-pairwise coprime moduli. In the example you provided, the moduli are not pairwise coprime, as the greatest common divisor (GCD) of 8 and 4 is 4, which is not equal to 1. Therefore, the code should not be used to compute the result for such input.

I have updated the code to handle such cases and throw an exception stating that Moduli must be pairwise coprime.

HimeshKohad (commented):
Resolved recommended changes in the code

vil02 (changes requested):
Before you commit your code, run `clang-format` on it. You can do it by executing the following command in your terminal (_while being_ in the directory of this project):
```bash
clang-format -i --style=file path_to_your_file.cpp
```
The `--style=file` makes it to use [.clang-format](https://github.com/TheAlgorithms/C-Plus-Plus/blob/6daf7015e31d079ae5896bb20719e46ccf8dd821/.clang-format) as the _style_ configuration.

vil02:
@HimeshKohad this PR is a very nice contribution. Could you please fix the missing details? It would be a pity if this would be closed without merging.

PR #2524, opened 2023-08-24
Fix: fit ```Eratosthenes.cpp``` to guidelines

ewd00010:
#### Description of Change
chores: removed code from main, moved time check to tests, small changes to comments and documentation

#### Checklist

Notes: do we want to keep the time diff calculation?

### DIFF
--- a/math/eratosthenes.cpp
+++ b/math/eratosthenes.cpp
@@ -1,15 +1,14 @@
 /**
  * @file
- * @brief [The Sieve of
+ * @brief A way to find prime numbers up to the specified number, [The Sieve of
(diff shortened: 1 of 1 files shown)


math/eratosthenes.cpp
```diff
@@ -75,44 +88,14 @@ static void test() {
     assert(primes[8] == false);
     assert(primes[9] == false);
 
-    std::cout << "All tests have successfully passed!\n";
```
realstealthninja, line 78: \n is apparently better than std::endl it increases performance slightly but this is also good

ewd00010: yea, its not worth changing in this one, but i'll do \n in the future!

PR #2542, opened 2023-09-26
docs: Update sum_of_digits.cpp documentation

dylancusson:
#### Description of Change
Updates to the documentation, tests, and algorithm of sum_of_digits.cpp
Edited documentation for clarity.
Tidied up tests from multiple functions into one that handles several asserts.
Updated algorithm to retain signed value, ex: sum_of_digits(-123) == -6 instead of 6.

#### Checklist

Notes: Updated documentation, tests, and algorithm for sum_of_digits.cpp


math/sum_of_digits.cpp
```diff
@@ -1,72 +1,74 @@
...
- * Next, the algorithm extract the last digit of num by dividing by 10
- * and extracting the remainder and this is added to the sum.
- * The number is then divided by 10 to remove the last digit.
- * This loop continues until num becomes 0.
  */
 int sum_of_digits(int num) {
-    // If num is negative then negative sign is neglected.
+    /*
+     * First the algorithm checks whether the number is negative or positive,
+     * and sets the sign accordingly.
```
realstealthninja, line 26: the answer should always be positive no?

realstealthninja (changes requested):

math/sum_of_digits.cpp
dylancusson: https://en.wikipedia.org/wiki/Digit_sum
Was based on the wikipedia article which allows for negative digits, though I suppose with that interpretation then the whole `sign` and `num < 0` business is unnecessary.

PR #2574, opened 2023-10-01
factorial_iterative.cpp

tarunkeshukumar:
#### Description of Change
Created iterative version of finding factorial of a number.


CarlosZoft:
iterative factorial already exists 

you can see, here :
[iteractive factorial](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/math/iterative_factorial.cpp)

@tarunkeshukumar

PR #2592, opened 2023-10-04
Add Banker's algorithm implementation in C++.

asinghchandel082:
Add Banker's algorithm implementation in C++.

This pull request adds an implementation of the Banker's algorithm in C++. The Banker's algorithm is a resource allocation algorithm that can be used to prevent deadlocks.

The implementation in this pull request is based on the following pseudocode:

initialize the available resources
initialize the maximum demand of each process

while all processes are not finished:
  find a process that is not finished and whose need is less than or equal to the available resources
  allocate the resources to the process
  remove the process from the list of processes

if all processes are finished:
  the system is in a safe state
else:
  the system is in an unsafe state

The implementation in this pull request also includes a function to grant a resource request to a process. This function checks if the request is within the process's maximum demand and if the system is still in a safe state after granting the request. If both conditions are met, the request is granted.

I have tested the implementation and it seems to be working correctly.

I would appreciate it if you would review my pull request and let me know if you have any feedback.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -87,5 +87,6 @@
     "xtree": "cpp",
     "xutility": "cpp",
     "climits": "cpp"
-  }
+  },
+  "C_Cpp.errorSquiggles": "disabled"
 }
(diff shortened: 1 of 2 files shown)


realstealthninja:
Please use the given PR format.

realstealthninja (changes requested):
Documentation is a MUST for all pull requests as such please follow contribution.md for the proper documentation standard

PR #2607, opened 2023-10-08
fix: overflow when searching the middle of a segment

marishok:
Fixed finding the middle of the segment. Now if "a" and "b" fit into the double type, then the middle too. In the previous implementation, there could be an exit beyond the size of double.

### DIFF
--- a/numerical_methods/bisection_method.cpp
+++ b/numerical_methods/bisection_method.cpp
@@ -55,7 +55,7 @@ int main() {
 
     // start iterations
     for (i = 0; i < MAX_ITERATIONS; i++) {
-        x = (a + b) / 2;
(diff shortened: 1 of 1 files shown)


numerical_methods/bisection_method.cpp
```diff
@@ -55,7 +55,7 @@ int main() {
 
     // start iterations
     for (i = 0; i < MAX_ITERATIONS; i++) {
-        x = (a + b) / 2;
+        x = a + (b - a) / 2;
```
realstealthninja, line 58: doesn't this just result in b?
a + b -a = b

realstealthninja (changes requested):

PR #2661, opened 2023-11-18
提交了#2640

hello-dys:
1.Activity Selection Problem
2.Job Sequencing Problem
3.Huffman Coding
4.Huffman Decoding
5.Water Connection Problem

### DIFF
--- a/greedy_algorithms/activity choice.cpp
+++ b/greedy_algorithms/activity choice.cpp
@@ -1,0 +1,31 @@
+#include<bits/stdc++.h>
+using namespace std;
+
+struct Activity {
+    int start, finish;
+};
+
+bool activityCompare(Activity s1, Activity s2) {
+    return (s1.finish < s2.finish);
+}
+
+void printMaxActivities(Activity arr[], int n) {
+    sort(arr, arr + n, activityCompare);
+
+    int i = 0;
+    cout << "(" << arr[i].start << ", " << arr[i].finish << "), ";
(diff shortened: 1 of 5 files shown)


greedy_algorithms/activity choice.cpp
```diff
@@ -0,0 +1,31 @@
+#include<bits/stdc++.h>
```
realstealthninja, line 1: Should not use this header ✖️

greedy_algorithms/activity choice.cpp
```diff
@@ -0,0 +1,31 @@
...
+using namespace std;
```
realstealthninja, line 2: Using namespace std is bad practice

greedy_algorithms/job sequencing.cpp
```diff
@@ -0,0 +1,37 @@
...
+        if (!visited[adj[v][i]])
+            topologicalSortUtil(adj[v][i], visited, Stack, adj);
+
+    Stack.push(v);
+}
+
+void topologicalSort(vector<int> adj[], int V) {
+    stack<int> Stack;
+    bool* visited = new bool[V];
+    for (int i = 0; i < V; i++)
```
realstealthninja, line 21: Must use {} for everything instead of relying on indentation

realstealthninja (changes requested):
Please seperate these files into multiple prs

PR #2670, opened 2024-01-03
adding missing cstdint for uint32_t

yahya-mohammed07:
#### Description of Change

Tested on linux machine it turns out the adding `#include <cstdint>` in necessary . Tested sieve_of_eratosthenes.cpp file only

### DIFF
--- a/math/sieve_of_eratosthenes.cpp
+++ b/math/sieve_of_eratosthenes.cpp
@@ -10,7 +10,7 @@
  *
  * @see primes_up_to_billion.cpp prime_numbers.cpp
  */
-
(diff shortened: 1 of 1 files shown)


realstealthninja:
not required as it always compiles in the github actions. i.e no errors are present when compiled by the github action. probably a misconfigured cpp installation on your part

yahya-mohammed07:
> not required as it always compiles in the github actions. i.e no errors are present when compiled by the github action. probably a misconfigured cpp installation on your part

check this [cpp_reference](https://en.cppreference.com/w/cpp/types/integer)

PR #2841, opened 2024-10-15
doc, fix: typo in merge sort and use vector instead of raw array 

pipinstallaadit:
1. Typographical Error in Doxygen Comment

In the Doxygen comment at the beginning of the file, "Merege Sort" should be corrected to "Merge Sort".

2. Merging Logic

In the merge function, the merging logic can lead to out-of-bounds access if both sub-arrays have been completely traversed. The condition in the while loop should ensure that you only access elements of L and R if they are within bounds:

3. Memory Management

Using new and delete[] for the temporary arrays (L and R) is fine, but you can also use std::vector from the C++ Standard Library to simplify memory management:

4. Input Validation

You should consider adding input validation when reading the number of elements and the actual elements to avoid undefined behavior.

5. Main Function Improvements

You can also enhance the main function by encapsulating the input logic in a separate function, improving readability.

#### Checklist

- [ x] Added description of change
- [ x] Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#New-File-Name-guidelines)
- [ x] Added tests and example, test must pass

Notes:


sorting/merge_sort.cpp
```diff
@@ -54,8 +53,17 @@ void merge(int *arr, int l, int m, int r) {
         k++;
     }
 
-    delete[] L;
-    delete[] R;
+    while (i < n1) {
+        arr[k] = L[i];
```
realstealthninja, line 57: what is this doing?

realstealthninja (changes requested):

sorting/merge_sort.cpp
pipinstallaadit: These loops are necessary because, in some cases, either L or R may still contain elements after the main merging loop has finished. These remaining elements are already sorted within their respective arrays, so they can be directly copied into arr.

PR #2914, opened 2025-01-11
DFS algorithm

nikolassc:
This implementation of the Depth First Search (DFS) algorithm provides a robust and flexible solution for graph traversal. It uses an adjacency list representation of the graph, implemented with std::unordered_map<int, std::vector<int>>, allowing it to efficiently handle sparse and large graphs. The algorithm is implemented recursively, visiting each node and its neighbors while tracking visited nodes using a std::unordered_set to prevent revisiting nodes during the traversal. This ensures it can handle cyclic graphs without entering infinite loops.

Additionally, the implementation includes error handling to address invalid starting nodes. If the provided starting node is not present in the graph, the function throws a std::invalid_argument exception with a clear error message, making it resilient against incorrect inputs. The algorithm has been tested on various graph scenarios, including connected, disconnected, and cyclic graphs, and demonstrates correct and predictable behavior in all cases.

The DFS algorithm is efficient, with a time complexity of O(V + E), where V is the number of vertices and E is the number of edges, and a space complexity of O(V) due to the use of the visited set and recursion stack. This implementation is designed to be clear, flexible, and scalable, suitable for a wide range of graph-related applications.

### DIFF
--- a/search/depth_first_search.cpp
+++ b/search/depth_first_search.cpp
@@ -1,0 +1,160 @@
+/******************************************************************************
+ * @file
+ * @brief Depth First Search (DFS) Algorithm Implementation
+ * @details
+ * Depth First Search (DFS) is an algorithm used for traversing or searching
+ * tree or graph data structures. The algorithm starts at a given starting node
+ * and explores as far as possible along each branch before backtracking.
+ *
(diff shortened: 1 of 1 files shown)


nikolassc:
@Panquesito7 
@realstealthninja

PR #2970, opened 2025-07-19
Add count distinct primes from binary string algorithm

rudrakshtank:
#### Description of Change
Added `count_distinct_primes_from_binary_string.cpp` to the `other/` directory.

This program counts the number of distinct prime numbers that can be formed from a binary string using:

- Any permutation of characters (i.e., swapping)
- Changing '1's to '0's (but not the reverse)

It uses an efficient implementation of the Sieve of Eratosthenes for prime checking and bitmasking to explore all possible combinations.

Includes:
- Doxygen-style documentation
- A working `main()` with an example test case

[Contributors guide](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md)

---

#### Checklist

- [x] File name matches [File name guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#New-File-Name-guidelines)
- [x] Added test case/example in `main()`
- [x] Added Doxygen-style documentation
- [x] Comments/documentation added as needed
- [x] PR title follows [semantic commit guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#Commit-Guidelines)
- [x] Confirmed this is not a duplicate contribution
- [x] I acknowledge that all my contributions will be made under the project's license

---

#### Notes:
Efficient bitmask + sieve-based solution to find all valid prime numbers from binary strings using allowed transformations.


bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * Tags: Bit Manipulation, Prime Numbers, Combinatorics, Greedy, Bitmask
+ */
+
+#include <iostream>
+#include <vector>
+#include <unordered_set>
+#include <algorithm>
+
+const int MAX = 1e6;
```
realstealthninja, line 24: max cannot be a negative number, hence replace int with uint32_t

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,77 @@
...
+std::vector<bool> is_prime;
```
realstealthninja, line 25:
```suggestion
static std::vector<bool> is_prime;

```

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,77 @@
...
+
```
realstealthninja, line 26:
```suggestion

/**
 * @namespace bit_manipulation
 * @brief Bit manipulation algorithms
 */
namespace bit_manipulation {
```

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,77 @@
...
+            if (!seen.count(i)) {
+                cnt++;
+                seen.insert(i);
+            }
+        }
+    }
+
+    return cnt;
+}
+
```
realstealthninja, line 66:
```suggestion

} // bit_manipulation
```

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,77 @@
...
+int countPrimeBinaryStrings(const std::string &s) {
+    int n = s.length();
+    int k = std::count(s.begin(), s.end(), '1');
+    int cnt = 0;
+    int limit = 1 << n;
+
+    std::unordered_set<int> seen;
+
+    for (int i = 2; i < limit; i++) {
+        if (__builtin_popcount(i) <= k && is_prime[i]) {
```
realstealthninja, line 56: __builtin_popcount is not part of the STL library it only exists in GCC.
thus this should be replaced with `std::popcount()`
```suggestion
        if (std::popcount(i) <= k && is_prime[i]) {

```

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * Efficient solution using bit manipulation and Sieve of Eratosthenes.
+ *
+ * Tags: Bit Manipulation, Prime Numbers, Combinatorics, Greedy, Bitmask
+ */
+
+#include <iostream>
+#include <vector>
+#include <unordered_set>
+#include <algorithm>
```
realstealthninja, line 22:
```suggestion
#include <bit>             /// for std::popcount
#include <iostream>
#include <vector>
#include <unordered_set>
#include <algorithm>
```

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+/**
+ * @brief Main function to test the algorithm.
+ */
+int main() {
+    precomputePrimes();
+    std::string s;
+    std::cin >> s;
+    std::cout << countPrimeBinaryStrings(s) << std::endl;
```
realstealthninja, line 74:
```suggestion
```
none of this should be in the main function, Only a call to the tests function.

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,77 @@
...
+                cnt++;
+                seen.insert(i);
+            }
+        }
+    }
+
+    return cnt;
+}
+
+/**
```
realstealthninja, line 67: please add tests here

bit_manipulation/count_distinct_primes_from_binary_string.cpp
realstealthninja: Since you want to pre compute primes why not make this a constexpr? and assign in it directly? instead of a call from main. Also afaik vectors cant be used as constexpr in c++17 hence it might be preferred if you use std::array instead.

if any of this sounds confusing let me know I can explain or direct you to resources to learn about this feature of c++

rudrakshtank:
I have done with the suggested changes, please review the changes

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,85 @@
...
+        return cnt;
+    }
+    
+    void tests(){
+        precomputePrimes();
+        std::string s;
+        std::cin >> s;
+        std::cout << countPrimeBinaryStrings(s) << std::endl;
+    }
+
```
realstealthninja, line 77: tests should be outside the namespace and static, A test should be comparing a computed value to a known value using an assert statement

bit_manipulation/count_distinct_primes_from_binary_string.cpp
realstealthninja: you still havent done this
```suggestion
const uint32_t MAX = 1e6;

```

bit_manipulation/count_distinct_primes_from_binary_string.cpp
realstealthninja: please add documentation for all the above headers as shown in the example

realstealthninja (changes requested):

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -16,27 +16,31 @@
...
+#include <cassert>      ///< For assert-based testing
 
-const int MAX = 1e6;
+const uint32_t MAX = 1e6;  ///< Upper bound for prime sieve
 static std::vector<bool> is_prime;
 
 /**
  * @namespace bit_manipulation
- * @brief Bit manipulation algorithms
+ * @brief Bit manipulation algorithms and prime preprocessing
```
realstealthninja, line 31:
```suggestion
 * @brief Bit manipulation algorithms
```

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -67,17 +71,26 @@ namespace bit_manipulation {
...
+    // Example test cases
+    assert(countPrimeBinaryStrings("1") == 0);     // Only "1" -> not prime
+    assert(countPrimeBinaryStrings("11") > 0);     // Should form primes like 3
+    assert(countPrimeBinaryStrings("101") > 0);    // Can form primes like 5
+    assert(countPrimeBinaryStrings("000") == 0);   // No 1s -> no primes possible
+}
+
 /**
- * @brief Main function to test the algorithm.
+ * @brief Main function to run tests.
```
realstealthninja, line 93:
```suggestion
 * @brief Main function
 * @returns 0 on successful exit

```

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,98 @@
...
+ * @brief Static test function using assertions instead of I/O.
+ */
+static void tests() {
+    using namespace bit_manipulation;
+
+    precomputePrimes();
+
+    // Example test cases
+    assert(countPrimeBinaryStrings("1") == 0);     // Only "1" -> not prime
+    assert(countPrimeBinaryStrings("11") > 0);     // Should form primes like 3
```
realstealthninja, line 87: you need to compare against the actual value that should be forming, comparing this way against 0 is too general. here from my estimation 
01 not prime
10 prime
11 prime
so the answer should be three
you should be comparing against three

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,98 @@
...
+    assert(countPrimeBinaryStrings("101") > 0);    // Can form primes like 5
```
realstealthninja, line 88: Same thing here, saying any binary string which is non zero and non 1 can form primes above 0 is not a great test.
here i estimate
001 not prime
100 not prime
101 prime 

so the answer should be one, please compare against the actual number

realstealthninja (changes requested):

bit_manipulation/count_distinct_primes_from_binary_string.cpp
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * @namespace bit_manipulation
+ * @brief Bit manipulation algorithms
+ */
+namespace bit_manipulation {
+
+    /**
+     * @brief Precomputes all prime numbers up to MAX using Sieve of Eratosthenes.
+     */
+    void precomputePrimes() {
```
realstealthninja, line 38: Again would recommend using constexpr to do the computation at compile time

realstealthninja (changes requested):

PR #2997, opened 2025-09-10
Implemented LRU Cache using unordered_map + list

sachiny123:
This PR implements the Least Recently Used (LRU) Cache data structure in C++ using 
unordered_map and list for efficient O(1) access and updates. 

Features:
- get(key): Returns the value of the key if it exists, else returns -1.
- put(key, value): Inserts a new key-value pair or updates an existing key. 
  If the cache exceeds its capacity, the least recently used item is removed.
- Efficient O(1) time complexity for both get and put operations.
- Includes example usage in main() demonstrating functionality.

This implementation is practical for interview preparation and real-world applications 
where caching with eviction policy is required.


realstealthninja:
closing...

PR #3029, opened 2025-10-06
Fit binaryadd.cpp

Sujalgupta2p:
🧩 Add Binary — LeetCode Problem 67 (C++)

### 📘 Problem Description
This prolgram adds two binary strings and returns their sum as a binary string.

The logic simulates manual binary addition — moving from the least significant bit (rightmost) to the most significant bit (leftmost), while carrying over when necessary.  
This implementation focuses on **string manipulation**, **bit logic**, and **carry handling**, making it an excellent problem for improving algorithmic reasoning in C++.

---

### 🧠 How It Works
1. Start from the end of both binary strings `a` and `b`.
2. Add corresponding bits and a `carry` (initially `0`).
3. Append the result of `(sum % 2)` to a string.
4. Update `carry = sum / 2`.
5. Reverse the final string since addition proceeds right-to-left.

### DIFF
--- a/bit_manipulation/binaryadd.cpp
+++ b/bit_manipulation/binaryadd.cpp
@@ -1,0 +1,39 @@
+#include <bits/stdc++.h>
+using namespace std;
+
+class Solution {
+public:
+    string addBinary(string a, string b) {
+        string result = "";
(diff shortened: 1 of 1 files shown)


Sujalgupta2p:
Xyz

PR #3067, opened 2025-11-29
Add count_digits function with test cases

Lumowhisp:
This PR adds a small utility function to count the number of digits in an integer.
The implementation is simple and follows patterns used in other files in the math directory.
Basic test cases have been added using assert to ensure correctness.

#### Description of Change
	•	Added a new file count_digits.cpp under the math directory.
	•	Implemented count_digits(int n) to return the total number of digits in a given integer.
	•	Added a small test function using assert to verify expected outputs.
	•	Included brief comments for readability and consistency with other files.

#### Checklist
	•	Added description of change
	•	File name follows guidelines (count_digits.cpp)
	•	Added tests using assert
	•	Added minimal documentation for readability
	•	PR title follows semantic guidelines
	•	No duplicate implementation found
	•	I acknowledge that my contribution will be under the project’s license

#### Notes
Simple implementation of a digit-counting function with small test cases.
Suitable for the math section and consistent with existing examples.

### DIFF
--- a/math/count_digits.cpp
+++ b/math/count_digits.cpp
@@ -1,0 +1,27 @@
+#include <iostream>
+#include <cassert>
+
+// returns how many digits the number has
+int count_digits(int n) {
+    if (n == 0) return 1;
+
+    int c = 0;
+    while (n > 0) {
+        c++;
+        n /= 10;
+    }
+    return c;
+}
+
+void test() {
+    assert(count_digits(12345) == 5);
+    assert(count_digits(9) == 1);
+    assert(count_digits(1000) == 4);
+    assert(count_digits(0) == 1);
+}
+
+int main() {
+    test();
+    std::cout << "Tests passed\n";
+    return 0;
+}


math/count_digits.cpp
```diff
@@ -0,0 +1,27 @@
+#include <iostream>
+#include <cassert>
+
+// returns how many digits the number has
+int count_digits(int n) {
```
Rfontt: When the input is negative, the current implementation does not count its digits correctly because the loop only runs while `n > 0`.

Additionally, there’s a more subtle problem: the smallest signed integer `(INT_MIN = –2,147,483,648)` cannot be converted to a positive int using [abs()](https://cplusplus.com/reference/cstdlib/abs/), because its positive value is outside the valid range of a signed 32-bit integer.

**So, this leads to undefined behavior in C++.**

To avoid this issue, the value should be converted to an unsigned int before processing.
Unsigned integers do not store negative values, which eliminates the sign bit and allows representing the absolute value of every possible int, including `INT_MIN`.

**Example:**

`unsigned int x = (n < 0) ? -(unsigned int)n : (unsigned int)n;`

**This ensures:**

- Negative numbers are handled safely.
- INT_MIN is handled correctly.
- The digit-counting loop works for all integer inputs.

What do you think about this suggestion?

PR #3157, opened 2026-05-30
feat: add Segment Tree with Range Minimum Query using lazy propagation

binarymind-dev:
## Description
This PR adds a Segment Tree implementation focused on Range Minimum Query (RMQ) with lazy propagation for range updates.

## Changes
- Added `range_queries/segment_tree_rmq.cpp`
- Supports Range Minimum Query in O(logN)
- Supports Range Update with lazy propagation in O(logN)
- Includes self-test with assert statements

## Difference from existing segtree.cpp
- existing `segtree.cpp` supports Range Sum Query
- This implementation supports Range Minimum Query (RMQ)
- Class-based implementation with `range_queries` namespace

### DIFF
--- a/range_queries/segment_tree_rmq.cpp
+++ b/range_queries/segment_tree_rmq.cpp
@@ -1,0 +1,219 @@
+/**
+ * @file
+ * @brief   Implementation of [Segment Tree]
+ *          (https://en.wikipedia.org/wiki/Segment_tree) data structure
+ *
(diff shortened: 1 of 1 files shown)


binarymind-dev:
This PR is still relevant, just waiting for a code owner review. @Panquesito7 @realstealthninja could you please take a look when you get a chance

PR #3220, opened 2026-08-06
docs: drop the stray @param from the @returns lines in graham_scan_functions

darkdi:
#### Description of Change

Every `@returns` line in `geometry/graham_scan_functions.hpp` carries a stray `@param`:

```
 * @returns @param int distance between p1 and p2.
```

Doxygen reads that as a parameter named `int`, so the return value goes undocumented and a phantom parameter shows up in the generated page instead. Five occurrences, one per function in the file.

Two of them were also describing the wrong thing, and since I was editing the line anyway:

- `distSq` said "distance between p1 and p2" while both its `@brief` and its body return the *square* of the distance.
- `compare` said "distance between p1 and p2" too — copied from `distSq`. It is the `qsort` comparator and returns `-1` or `1`.

Also fixed "Conver Hull" to "convex hull" on the last one.

Comments only; no code touched, and the file stays inside one directory.

#### Checklist

Notes: comment-only change to one file, no behaviour or test impact.


realstealthninja (approved):

darkdi:
Still current. The macOS job dies in ciphers/hill_cipher.cpp, which this PR does not touch, and geometry/graham_scan_functions.hpp has not changed since 2024, so the approval still applies as is.

darkdi:
There is also a separate infrastructure failure in the approved-label job: it logs "Approval requirement is met, adding label" and then fails with "Resource not accessible by integration". Could a maintainer apply the approved label? The macOS failure mentioned above is in hill_cipher.cpp, outside this documentation change.

Issue #3203, opened 2026-07-14
[BUG] error: 'is_integral' cannot be specialized: Users are not allowed to specialize this standard library entity [-Winvalid-specialization]


yurivict:
### Description

```
In file included from /usr/ports/math/the-algorithms-c++/work/C-Plus-Plus-b9c118f/ciphers/uint256_t.hpp:13:
/usr/ports/math/the-algorithms-c++/work/C-Plus-Plus-b9c118f/ciphers/uint128_t.hpp:25:13: error: 'is_integral' cannot be specialized: Users are not allowed to specialize this standard library entity [-Winvalid-specialization]
   25 | struct std::is_integral<uint128_t> : std::true_type {};
      |             ^
/usr/include/c++/v1/__type_traits/is_integral.h:25:8: note: marked '_Clang::no_specializations' here
   25 | struct _LIBCPP_NO_SPECIALIZATIONS is_integral : _BoolConstant<__is_integral(_Tp)> {};
      |        ^
/usr/include/c++/v1/__config:1167:9: note: expanded from macro '_LIBCPP_NO_SPECIALIZATIONS'
 1167 |       [[_Clang::__no_specializations__("Users are not allowed to specialize this standard library entity")]]
      |         ^
In file included from /usr/ports/math/the-algorithms-c++/work/C-Plus-Plus-b9c118f/ciphers/elliptic_curve_key_exchange.cpp:27:
In file included from /usr/ports/math/the-algorithms-c++/work/C-Plus-Plus-b9c118f/ciphers/uint256_t.hpp:13:
/usr/ports/math/the-algorithms-c++/work/C-Plus-Plus-b9c118f/ciphers/uint128_t.hpp:27:13: error: 'is_arithmetic' cannot be specialized: Users are not allowed to specialize this standard library entity [-Winvalid-specialization]
   27 | struct std::is_arithmetic<uint128_t> : std::true_type {};
      |             ^
```

clang-21.1.8
FreeBSD 15.1 

### Expected behavior

n/a

### Actual behavior

n/a

### Steps to reproduce

compile

### Context

n/a

### Additional information

n/a

piyushadke7:
Hi! I'd like to work on this issue. Could you please assign it to me?
Thank you!

AashiSrivastava411:
[...]

PR #3230, opened 2026-09-12
Fix: [BUG] error: 'is_integral' cannot be specialized: Users are not allowe

goransh-walia:
This PR addresses #3203.

[BUG] error: 'is_integral' cannot be specialized: Users are not allowed to specialize this standard library entity [-Winvalid-specialization]

Generated with AI assistance and validated against the original
file before submission (syntax check + change-scope check).
Please review carefully — happy to adjust based on feedback.


bbmvm59ccc-ai (approved):

realstealthninja:
This is a very bad solution, merely kicking it forward without addressing the issue at hand.

P.S. the maintainers are discussing the ban of LLM assisted commits to the repository

realstealthninja (changes requested):
See comment above. Preferably close this pr. And open another actually fixing the issue.

